PR #710, opened 2017-08-22
delete ex.printStackTrace() code

powerYao:


apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/ClassLoaderUtil.java
```diff
@@ -37,7 +37,6 @@
       }
     } catch (Throwable ex) {
       classPath = System.getProperty("user.dir");
-      ex.printStackTrace();
```
nobodyiam: how about "logger.warn("Failed to locate class path, fallback to user.dir: {}", classPath, ex)"?

powerYao: ok

PR #784, opened 2017-10-10
use java8 api replace guava joining List to String

Acceml:


apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/controller/ConfigController.java
```diff
@@ -113,8 +111,8 @@ public ApolloConfig queryConfig(@PathVariable String appId, @PathVariable String
 
     auditReleases(appId, clusterName, dataCenter, clientIp, releases);
 
-    String mergedReleaseKey = FluentIterable.from(releases).transform(
-        input -> input.getReleaseKey()).join(STRING_JOINER);
+    String mergedReleaseKey = releases.stream().map(Release::getReleaseKey)
+            .collect(Collectors.joining(ConfigConsts.CLUSTER_NAMESPACE_SEPARATOR));
```
nobodyiam: Looks like `Collectors.joining(ConfigConsts.CLUSTER_NAMESPACE_SEPARATOR)` will create a new object everytime, which will bring some performance overhead, is it possible to cache it?

If we would like to cache `Collectors.joining(ConfigConsts.CLUSTER_NAMESPACE_SEPARATOR)`, then we need to make sure it is thread-safe, would you please help to give it a check? Thanks!

Acceml: 1.There is no performance overhead.`Collectors.joining` ,used in lamda expression，is a high level abstract of `StringJoiner` and `StringBuilder`.
```
    public static Collector<CharSequence, ?, String> joining(CharSequence delimiter,
                                                             CharSequence prefix,
                                                             CharSequence suffix) {
        return new CollectorImpl<>(
                () -> new StringJoiner(delimiter, prefix, suffix),
                StringJoiner::add, StringJoiner::merge,
                StringJoiner::toString, CH_NOID);
    }
```
```
    public StringJoiner add(CharSequence newElement) {
        prepareBuilder().append(newElement);
        return this;
    }
```
2.There seems no need to care the thread-safe.The `Spring MVC`, providing Http service, has aggregated the `Servlet containers`.When a Http request arrives, the container picks a Thread from the pool and executes.So, it is a purely local thread without accessing common data.Furthermore, We don't use `releases.parallelStream().map` in the code.

3.The `Collectors.joining` is not thread-safe, and it is safe used in lamda luckily.
The top answer of [Parallel streams, collectors and thread safety
](https://stackoverflow.com/questions/22350288/parallel-streams-collectors-and-thread-safety)
and the [offcial doc](https://docs.oracle.com/javase/8/docs/api/java/util/stream/IntStream.html) has mentioned:
> collect operations can be parallelized without requiring additional synchronization.

4. Idea recommends to use java8 api:
qq 20171012103306

nobodyiam: Thanks for the research! What I mean is `Collectors.joining(ConfigConsts.CLUSTER_NAMESPACE_SEPARATOR)` will create 2 objects everytime it is called, `new CollectorImpl` and `new StringJoiner`.

Considering the case in large deployment environments, there maybe thousands of calls to this method per second, which might create some performance overhead? How do you think?

Acceml: I think there is a disagreement about execution of streaming. The `collect ` is a  terminal operation in streaming, allowing for reuse of collection strategies.
Collectors provides a series of strategies.
image
```
    public static <T>
    Collector<T, ?, List<T>> toList() {
        return new CollectorImpl<>((Supplier<List<T>>) ArrayList::new, List::add,
                                   (left, right) -> { left.addAll(right); return left; },
                                   CH_ID);
    }
```
Just as `.collect(Collectors.toList())` wouldn't new A list every time, `.collect(Collectors.joining())` would not `new CollectorImpl` and `new StringJoiner`.

nobodyiam: It has nothing to do with streaming.

If you call `queryConfig`, a new `CollectorImpl` and a new `StringJoiner` will be created. 

If you call `queryConfig` again, another `CollectorImpl` and `StringJoiner` will be created, isn't it?

nobodyiam: I checked the implementation of `CollectorImpl` and `StringJoiner` and found out the overhead should be very small, so I think it's ok.

PR #814, opened 2017-11-03
Add callback to defferdResult before putting to the map

xiaoerlyl:


xiaoerlyl:
When i test config-service with 10k concurrency, the memory usage is very high. I guess maybe somewhere memory leak.

nobodyiam:
OK. Did the change work?

If not, I suggest you use java flight recorder to record the memory usage.

We have done such test before, which turns out most of the memory were used by AntPathMatcher.
[More info](http://lepdou.github.io/blogs/apollogc/gc.html)

xiaoerlyl:
It does not work. 
Thanks very much for your suggestion, it really helps.

PR #841, opened 2017-11-23
support better rollback.

xiaoerlyl:
The current rollback does not work well, because the changes after the release  were kept in the database. I think we should delete these changes and rollback the configurations to a clean version.


xiaoerlyl:
1. This pull request still uses soft delete.
2. The greatest need for this pull request is that i can not rollback the deleted item. In the next release, the deleted item has to been deleted again.
3. You are right, the item info in release configurations is not enough to rollback to an old version clearly.

PR #845, opened 2017-11-27
clear http response inputstream and errorstream to reuse connection.

xiaoerlyl:
According to this official article:  [https://docs.oracle.com/javase/7/docs/technotes/guides/net/http-keepalive.html](https://docs.oracle.com/javase/7/docs/technotes/guides/net/http-keepalive.html), we should clear http response inputstream and errorstream to reuse this connection.


xiaoerlyl:
sorry, I missed some cases. It is when server status code is more than 400, you will get an excepion if you try to open input stream. I will fix it and open another pull request. 
1. you should test scenarios  when server response code is more than 400.
2. In normal cases, you will get null if you try to get error inputstream, and it is ok.
3.  It is when server status code is more than 400, you will get an excepion if you try to open input stream

PR #888, opened 2017-12-18
添加了一个新的注解EnableAutoResfresh

tonyDJB:
这个注解的作用是在变量上声明需要框架自动刷新，这样就不需要自己再写onChange事件，对简单拆箱的变量能省不少事


tonyDJB:
干的事情就是添加了一个名为EnableAutoResfresh 的注解
主要的逻辑在ApolloAnnotationProcessor.java

nobodyiam:
好的，这几天比较忙，周末看下。

apollo-client/src/main/java/com/ctrip/framework/apollo/spring/annotation/ApolloAnnotationProcessor.java
```diff
@@ -1,89 +1,138 @@
...
  */
 public class ApolloAnnotationProcessor implements BeanPostProcessor, PriorityOrdered {
-  @Override
-  public Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
-    Class clazz = bean.getClass();
-    processFields(bean, clazz.getDeclaredFields());
-    processMethods(bean, clazz.getDeclaredMethods());
-    return bean;
-  }
+    @Override
```
nobodyiam: Please use Apollo's [code style](https://github.com/ctripcorp/apollo/tree/master/apollo-buildtools/style) so that we won't have code style differences.

apollo-client/src/main/java/com/ctrip/framework/apollo/spring/annotation/ApolloAnnotationProcessor.java
```diff
@@ -1,89 +1,138 @@
...
+            }
+            String namespace = annotation.value();
+            final Config config = ConfigService.getConfig(namespace);
+            config.addAutoChangeListener(new AutoConfigChangeListener() {
+                @Override
+                public void autoChange(ConfigChangeEvent changeEvent) {
+                    ReflectionUtils.makeAccessible(field);
+                    //If it is the processing of complex data types, I suggest not to do AutoRefresh here or to handle it yourself
+                    //add by 258737400@qq.com 2017-12-18
+                    String type = field.getGenericType().toString();
```
nobodyiam: Remember we should always use Spring's `environment.getProperty()` to get the property value because the key might be overridden in a different hierarchy.

apollo-client/src/main/java/com/ctrip/framework/apollo/AutoConfigChangeListener.java
```diff
@@ -0,0 +1,17 @@
+package com.ctrip.framework.apollo;
+
+        import com.ctrip.framework.apollo.model.ConfigChangeEvent;
+
+/**
+ * @author tony Jiang(258737400@qq.com)
+ */
+public interface AutoConfigChangeListener {
```
nobodyiam: I don't think we need a new list of AutoConfigChangeListeners. If we want to make sure some listener is executed first, we could create a a new interface such as `PriorityOrderedConfigChangeListener` that extends ConfigChangeListener and allow the user to provide its order?

apollo-client/src/main/java/com/ctrip/framework/apollo/spring/annotation/EnableAutoResfresh.java
```diff
@@ -0,0 +1,27 @@
...
+ * @author Tony Jiang(258737400@qq.com)
+ */
+@Target({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.ANNOTATION_TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface EnableAutoResfresh {
+  /**
+   * Apollo namespace for the config, if not specified then default to application
+   */
+  String value() default ConfigConsts.NAMESPACE_APPLICATION;
```
nobodyiam: I think we don't require the use to specify the namespace, since spring property source has hierarchies, which means one key could exist in different namespaces.

apollo-demo/src/main/java/com/ctrip/framework/apollo/demo/spring/autoRefreshDemo/AutoRefreshDemo.java
```diff
@@ -0,0 +1,58 @@
...
+import com.ctrip.framework.apollo.demo.spring.common.bean.AutoRefreshBean;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.support.ClassPathXmlApplicationContext;
+
+import java.util.Scanner;
+
+/**
+ * @author tony jiang(258737400@qq.com)
+ */
+public class AutoRefreshDemo {
```
nobodyiam: This change impacts a lot, so we should have more tests to cover the value change scenarios.

apollo-client/src/main/java/com/ctrip/framework/apollo/AutoConfigChangeListener.java
tonyDJB: My idea is to provide a flexible refresh mode for users:

1. based on  ApolloConfigChangeListener annotation, let users write their own refresh code

2. based on AutoConfigChangeListener annotation, automatic assembly of data for simple data types

In addition, with regard to #898, I do not agree with this coercion, which forces all configuration variables to be refreshed automatically, thus losing the flexibility of the user.

Above all, I think it's better to provide an independent <AutoConfigChangeListener> annotation, because this is the two separate listener behavior, and there is no distinction between execution sequence.

apollo-client/src/main/java/com/ctrip/framework/apollo/spring/annotation/ApolloAnnotationProcessor.java
tonyDJB: Thanks for your comment, I will reformat my code.

apollo-client/src/main/java/com/ctrip/framework/apollo/spring/annotation/EnableAutoResfresh.java
tonyDJB: Thank you for your comment.

If namespace is not named, does the key of the same name in different namespace be confused with each other？

apollo-demo/src/main/java/com/ctrip/framework/apollo/demo/spring/autoRefreshDemo/AutoRefreshDemo.java
tonyDJB: Thank you for your comment.

I'm going to improve the test code.

apollo-client/src/main/java/com/ctrip/framework/apollo/spring/annotation/ApolloAnnotationProcessor.java
tonyDJB: What do you mean by modifying the code here?

tonyDJB:
Thank you for your advice in your busy schedule

I submitted comment under your question.

nobodyiam:
@tonyDJB 

非常感谢你提供的实现方式，考虑再三后，我们觉得可能对大部分用户而言，默认开启自动更新是有意义的，所以我们后来基于 #898 的提交做了一些改进，支持了对于各种情况下的placeholder值自动更新，具体细节可以参考 #972，再次感谢！

PR #1315, opened 2018-08-01
Refactor meta server related logic so that apollo-core is common enough to go maven central repository.

nobodyiam:
# Purpose

To refactor meta server related logic so that apollo-core is common enough to go maven centrl repository.

# Brief changes

1. Remove apollo-env.properties from apollo-core, so that it's possible
to deploy apollo-client to maven central repository.
2. Add MetaServerProvider spi so that users could customize how they
find apollo meta servers.
3. Refactor env related utils so that Env.UNKNOWN is returned for
invalid env string instead of null.

# Impact

From 1.0.0 version, apollo-core.jar won't assemble `apollo-env.properties` anymore, so that we could freely deploy apollo client jars to maven central repository, which means a lot convenience for users to access apollo client jars.

However, for existing users, they can't upgrade apollo client jars to 1.0.0 version directly.

# How to config meta server address?

Previously, the meta server address is assembled into apollo-core.jar during build time, so that end user could use apollo with zero config (assume the `env` is configured correctly).

With 1.0.0 version, the meta server address could be configured via the following ways, ordered by priority:

1. JVM system property `apollo.meta`
    * -Dapollo.meta=http://config-service-url
    * System.setProperty("apollo.meta", "http://config-service-url");
2. OS environment variable `APOLLO_META`
    * APOLLO_META=http://config-service-url
    * Please note that Linux os environment variable could not contain dot, so we use underscore instead
3. `apollo.meta` property from `server.properties`
4. `apollo.meta` property from `app.properties`
5. JVM system property `${env}_meta`
    * e.g. if current env is `dev`, then users could config `-Ddev_meta=http://config-service-url`
6. `${env.meta}` from `apollo-env.properties`
    * Users could still put `apollo-env.properties` into classpath or spring boot config folder and apollo will read from it
    * apollo-portal will still use this way to configure meta server addresses

# How to customize meta server address locating logic?

With 1.0.0 version, a MetaServerProvider spi is provided, users could inject their own MetaServerProvider to customize the meta server address locating logic.

Since we use the typical [Java Service Loader pattern](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html), it's quite easy to do so.

One thing to note is apollo will iterate through all the MetaServerProvider and will stop when any MetaServerProvider gives the meta server address, so users should be careful about what the order value is. The rule is the smaller order takes the higher priority, so the MetaServerProvider with order 0 will be in front of the MetaServerProvider with order 1.

# How can existing users upgrade to 1.0.0?

1. If existing users currently pass `env` and `${env}_meta` at runtime, they could continue to do so with no impact.
2. If existing users currently place `apollo-env.properties` in their classpath, they could continue to do so with no impact.
3. If existing users assemble meta server addresses into `apollo-core.jar`, then there are several solutions:
    * Create a `xx-company-apollo-client` jar with `apollo-env.properties` in classpath and dependens on apollo-client. Then ask the users to depend on `xx-company-apollo-client` jar instead of `apollo-client`. This might be the most tranparent way.
    * Ask the ops to configure `apollo.meta` in `server.properties` or os env `APOLLO_META`
    * Ask the user to configure `apollo.meta` by `-Dapollo.meta=http://config-service-url` or app.properties manually
    * Ask the user to put the `apollo-env.properties` in their projects' classpath


JaredTan95:
it's awesome! looking forward to 1.0.0. 😄

PR #1333, opened 2018-08-06
fix load remote config repo by wrong namespace

ChaochengRen:
case:
`
Config xmlConfigFile = ConfigService.getConfigFile("application", ConfigFileFormat.XML);
// will load namespace “application.xml” instead of "application"
`


nobodyiam:
In this case, loading `application.xml` is the right logic, which means it's not an issue.

nobodyiam:
closed for now

PR #1363, opened 2018-08-13
unify portal logout url to /user/logout

nobodyiam:


nobodyiam:
Fix #1359 when users customized the authentication

PR #1412, opened 2018-08-28
bugfix:modify string in its iteration

ChaochengRen:
Modifying string in its iteration, which will increase iteration times massively.


nobodyiam:
Interesting...

PR #1476, opened 2018-09-12
allow public namespaces created with no org prefix

nobodyiam:
# Purpose

To allow the user create public namespaces with no org prefix

# Brief changes

1. Added a `append org prefix checkbox` in the namespace creation page
2. Will check public namespace name **_global_** uniqueness before creation
    * global uniqueness means the name is not used as any public namespace as well as any private namespace

# Effect

#### 1. when `append org prefix checkbox` is checked (default behavior)

image

#### 2. when `append org prefix checkbox` is unchecked

image


chickenlj:
Awesome, I think this can make the config model more flexible. Sometimes, people may need global namespace.

PR #1494, opened 2018-09-18
support rolling clean up log file,based size and time policy.

JaredTan95:


nobodyiam:
I upgraded apollo's spring boot to 2.0 recently, so that we could use spring boot's built-in feature (`logging.file.max-size` and `logging.file.max-history`) to do the job instead, see #1530

PR #1586, opened 2018-10-19
Block delete default cluster

zouyx:
修复
* 演示环境删除某项目default集群，回到界面报错  Ref：#1571


nobodyiam:
Thanks, will take a look soon.

PR #1837, opened 2019-01-05
打包到docker脚本

silencecorner:
写了个简单脚本从release中下载编译好的文件，然后创建镜像到本地仓库


scripts/apollo-on-kubernetes/build-images.sh
```diff
@@ -0,0 +1,251 @@
...
+
+# envoriment check
+if [[ $(which docker|wc -l) -ne 1 ]] && [[ $(which wget|wc -l) -ne 1 ]] && [[ $(which unzip|wc -l) -ne 1 ]];then
+    echo "docker、wget and unzip must be installed!"
+else     
+    echo "envoriment checked success"
+fi
+
+getopt --test 
+if [ "$?" != "4" ];then
```
nobodyiam: I have installed getopt via `brew install gnu-getopt` but it seems not working, the value of `$?` is always 0.

when I run `getopt --test`, the result is `--`, it there something wrong with my installation?

silencecorner: output `--`  is no problem! `$?` is always 0 after a correct execution
thanks for your kind help!

silencecorner:
😅

PR #1839, opened 2019-01-05
enhancement: use constructor injection instead of field injection

kezhenxu94:
This patch uses Constructor Injection to avoid [the drawbacks of Field Injection](
https://www.google.com.hk/search?client=safari&rls=en&q=spring+field+injection&ie=UTF-8&oe=UTF-8&gws_rd=cr). What's more, Spring Documentation also recommends using Constructor Injection [here](https://docs.spring.io/spring/docs/4.2.x/spring-framework-reference/html/beans.html#beans-setter-injection)

Here are some of the drawbacks of Field Injection listed in [this article](https://blog.marcnuri.com/field-injection-is-not-recommended/):

- Disallows immutable field declaration

- Eases single responsibility principle violation

- Tightly coupled with dependency injection container

- Hidden dependencies


kezhenxu94:
There are some Field Injections left and some workarounds such as `@Lazy` annotation, but it needs refactoring if we want to replace them with Constructor Injection, and I'll do that if it's acceptable to you @nobodyiam , if any further changes should be done, please let me know. Thanks

nobodyiam:
@kezhenxu94 Thanks!

I can see using constructor injection makes the program more robust and more friendly to tests. However, I'm curious about how did you manage to modify so many places, by hand or by some tool?

I also notice there are some places marked with `@Lazy` and some places not, is there any simple rule to distinguish the different usage?

kezhenxu94:
> I can see using constructor injection makes the program more robust and more friendly to tests. However, I'm curious about how did you manage to modify so many places, by hand or by some tool?

Modifying those places with modern IDE is a easy task, I'm using Intellij.

> I also notice there are some places marked with `@Lazy` and some places not, is there any simple rule to distinguish the different usage?

The rule is that when the dependencies form a circle, such as `A` depends on `B` and `B` depends on `A`, then the dependencies must be annotated as `@Lazy` or exceptions will be thrown

nobodyiam:
@kezhenxu94 

Thanks for the explanation. Regarding the cyclic reference, how did you identify such case, manually or by some tool?

kezhenxu94:
> @kezhenxu94
> 
> Thanks for the explanation. Regarding the cyclic reference, how did you identify such case, manually or by some tool?

It's hard to identify that, though there are tools to detect that in [this thread](https://stackoverflow.com/a/8416268/5200161), I identified by running the test cases and Spring will print out the cyclic denpendencies graph.

nobodyiam:
@kezhenxu94 

Thanks! Will take a detailed look soon!

PR #1851, opened 2019-01-09
Minor: fix unit test. run all tests of this project will error, but single class or test will success.

heshengbang:
Minor: fix unit test. run all tests of this project will error, but single class or test will success.


nobodyiam:
Thanks, a similar patch was committed in [PR 1858](https://github.com/ctripcorp/apollo/pull/1858/files#diff-bd45ca7cbd7be20e074b9d93165ee01bR193)

PR #1863, opened 2019-01-12
feature: supports yml configuration format. fixes #653

### LINKED ISSUE #653: 有支持yaml的打算吗？
用习惯了yaml,当配置一些集合，数组的时候，感觉比properties好用多了

kezhenxu94:
This patch tries to fixes the issue #653 , though at this moment this patch only supports simple yml format and doesn't take complex situations into account such as array.

I open this PR in case that anyone is to do the same thing, and, if anyone has any advices after reviewing this commit, he/she can give constructive suggestions here and I'll update this patch.


apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/service/ItemService.java
```diff
@@ -61,8 +65,18 @@ public void updateConfigItemByText(NamespaceTextModel model) {
...
-    ConfigTextResolver resolver =
-        model.getFormat() == ConfigFileFormat.Properties ? propertyResolver : fileTextResolver;
+    ConfigTextResolver resolver = fileTextResolver;
+    switch (model.getFormat()) {
+      case Properties:
+        resolver = propertyResolver;
+        break;
+      case YML:
+      case YAML:
+        resolver = yamlResolver;
```
JaredTan95: I think `YML` can use this `yamlResolver`, too?

kezhenxu94: yes，they are essentially identical except the extension name

JaredTan95: > yes，they are essentially identical except the extension name

yeah, i mean that you can add `      case YML: resolver = yamlResolver;`  here.

kezhenxu94: There is no need to do so, we use fall back trick here, i.e. no break statement after YML, anyway, adding it explicitly may be a better practice.

nobodyiam:
To support yml, there are 2 possible solutions:

1. Use the same way as what we did for properties file and convert it to k-v when user submits it in portal ui
2. Store and publish the original yml texts and parse them in the client side

However, I'm not sure which one is better, i.e. easier to implement and easier to extend.

kezhenxu94:
> To support yml, there are 2 possible solutions:
> 
> 1. Use the same way as what we did for properties file and convert it to k-v when user submits it in portal ui
> 2. Store and publish the original yml texts and parse them in the client side
> 
> However, I'm not sure which one is better, i.e. easier to implement and easier to extend.

As I know, using solution 1 makes the key unreadable when the configuration is complex, which is exactly the situation where people prefer yml compared to other formats. The keys look like `akey[0].bkey`, `akey[1].bkey`  in arrays and `akey.bkey.ckey`, `akey,bkey.dkey` in objects. it would be worse when configuring object arrays.

Solution 2 would be flexible in client side and can be parsed into custom config bean.

So I prefer solution 1 ^_^

kezhenxu94:
Then I'll keep this PR opened and try another preferable solution(in my opinion), when it's done, we'll compare and pick one of them(if any of them is acceptable)

nobodyiam:
@kezhenxu94 

I'm working on #1944 to add yml support, which uses the solution `Store and publish the original yml texts and parse them in the client side`. Please take a look, thanks!

kezhenxu94:
@nobodyiam will take a look soon(perhaps later tonight or tomorrow)

PR #1869, opened 2019-01-14
Enhancement: validation

kezhenxu94:
After #1858 I found there are some validations using `com.ctrip.framework.apollo.common.utils.RequestPrecondition` that can also be replace with `javax.validation` api. Some extra unit tests are also add in this patch.


apollo-portal/pom.xml
```diff
@@ -31,6 +31,12 @@
 			<artifactId>h2</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<version>2.15.0</version>
```
nobodyiam: is there any new feature of mockito used?

kezhenxu94: Yes, that's another question I would ask. Why do we stick to mockito 1.10? The `spring-starter-test` specify the version 2.x, and in this patch, I create some integration tests that need to skip the authorization, see [SkipAuthorizationConfiguration](https://github.com/ctripcorp/apollo/pull/1869/files#diff-5311c9021e81de1f6cfb141fd4f1218a) for detail. The changes are not just for this patch, it should be useful in the future when we need to add more integration tests. If I've missed anything, please correct me. Thanks!

kezhenxu94: Clearly speaking, mockito 2.x allows us to mock beans that are to be injected into Spring container.

nobodyiam: Last time I tried to upgrade mocktio to 2.x, but met some errors in running all the tests

kezhenxu94: I know that, I've tried too, and I finally managed to eliminate all but one exception in `com.ctrip.framework.apollo.configservice.service.AppNamespaceServiceWithCacheTest#testAppNamespace`, and I don't find a way to solve this eventually, that's why I only upgraded the mockito version in module `apollo-portal`

nobodyiam: got it, I will try to fix that later (if ever possible)

nobodyiam: the `com.ctrip.framework.apollo.configservice.service.AppNamespaceServiceWithCacheTest#testAppNamespace` case should be fixed in #1895

PR #1883, opened 2019-01-17
Allow config value to be empty. fixes #1882

### LINKED ISSUE #1882: apollo能配置空字符串值吗？
有些环境的配置想设置为“空字符串值”，请问apollo能实现吗？
例如：
my.conf.val=

kezhenxu94:


kezhenxu94:
> Looks like this change doesn't work in my environment, the submit button is always disabled if value is empty. Is there other change required?

Yes I've confirmed that and I just update the patch, now it looks like this 

image

and after submitting:

image

PR #1944, opened 2019-02-03
add yaml support to apollo-client

### LINKED ISSUE #653: 有支持yaml的打算吗？

nobodyiam:
Fixes #653, #1113, #1321, #1488, #1585, #1698, 

1. support yaml config file transform to properties
2. support yaml config file injection to Spring


apollo-client/src/main/java/com/ctrip/framework/apollo/PropertiesCompatibleConfigFile.java
```diff
@@ -0,0 +1,16 @@
...
+ * Config files that are properties compatible, e.g. yaml
+ *
+ * @since 1.3.0
+ */
+public interface PropertiesCompatibleConfigFile extends ConfigFile {
+
+  /**
+   * @return the properties form of the config file or an empty properties if the content is invalid
+   */
+  Properties asProperties();
```
kezhenxu94: Is it really good to return an empty `properties` in case of the content is invalid? Then how do we tell that whether the content is empty or invalid. Isn't there such a situation that we need to do that?

nobodyiam: This is really a good question!

I think we need to give some thoughts into this...

kezhenxu94: My thought was throwing some exceptions like `MalformedYamlException` in case where the YAML document is invalid and returning empty `properties` in case where the YAML document is empty

nobodyiam: I think you are right, so `asProperties` will now throw a runtime exception if the content cannot be transformed to properties

PR #1969, opened 2019-02-15
Cleanup for 1.3.0

kezhenxu94:
- replace some guava API with new Java 1.7 native API
- replace Collections.sort with List.sort(Java 1.7)
- replace annonymous inner types with lambdas
- use try-with-resources where possible
- replace explicit types with diamond operators

### DIFF
--- a/apollo-common/src/main/java/com/ctrip/framework/apollo/common/utils/BeanUtils.java
+++ b/apollo-common/src/main/java/com/ctrip/framework/apollo/common/utils/BeanUtils.java
@@ -63,7 +63,7 @@ public class BeanUtils {
     final BeanWrapper src = new BeanWrapperImpl(source);
     PropertyDescriptor[] pds = src.getPropertyDescriptors();
 
-    Set<String> emptyNames = new HashSet<String>();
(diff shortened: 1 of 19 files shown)


apollo-common/src/main/java/com/ctrip/framework/apollo/common/utils/BeanUtils.java
```diff
@@ -1,232 +1,232 @@
...
-   * @param source
-   * @param target
-   */
-  public static void copyEntityProperties(Object source, Object target) {
-    org.springframework.beans.BeanUtils.copyProperties(source, target, COPY_IGNORED_PROPERTIES);
-  }
-
-  private static final String[] COPY_IGNORED_PROPERTIES = {"id", "dataChangeCreatedBy", "dataChangeCreatedTime", "dataChangeLastModifiedTime"};
-}
+package com.ctrip.framework.apollo.common.utils;
```
nobodyiam: what's changed in this file?

kezhenxu94: It's a matter of CRLF and is fixed

PR #1973, opened 2019-02-16
add view permission

SiouWang:
Only users with view right can see configuration.
By default an administrator has permissions to viewing configuration.


nobodyiam:
looks like the tests failed

PR #1996, opened 2019-02-21
[APOLLO-1966]: diff configurations among clusters. issue #1966

kezhenxu94:
related issue #1966


apollo-portal/src/main/resources/static/scripts/directive/namespace-panel-directive.js
```diff
@@ -701,6 +702,17 @@ function directive($window, toastr, AppUtil, EventManager, PermissionService, Na
                     + "&namespaceName=" + namespace.baseInfo.namespaceName;
             }
 
+            function goToDiffPage(namespace) {
+                if (!scope.lockCheck(namespace)) {
```
nobodyiam: Is this required?

kezhenxu94: @nobodyiam removed

nobodyiam:
Thanks! This feature looks nice, however, I think in the first step, we should automatically select the current cluster?

Suppose the app has 2 clusters, default and test, and if I clicked the `compare configurations` button in default cluster, then the default cluster should be selected automatically?

kezhenxu94:
> Thanks! This feature looks nice, however, I think in the first step, we should automatically select the current cluster?
> 
> Suppose the app has 2 clusters, default and test, and if I clicked the `compare configurations` button in default cluster, then the default cluster should be selected automatically?

Reasonable and user-friendly, I've updated as suggested

nobodyiam:
I'm also wondering whether the comparison of comments is valueable in all situations? 

Shall we provide an option like `compare comments` in the first step? I guess most of the time, what the users wants to compare may be the values instead of comments

kezhenxu94:
> I'm also wondering whether the comparison of comments is valueable in all situations?
> 
> Shall we provide an option like `compare comments` in the first step? I guess most of the time, what the users wants to compare may be the values instead of comments

Can we just put a toggle button in the step 2 page and users can toggle (whether showing the diff of comments) without going back to page 163.com?

nobodyiam:
add a toggle is a good idea

kezhenxu94:
@nobodyiam Changes are made as discussed, now it works like this:
image

image

apollo-portal/src/main/resources/static/config/diff.html
```diff
@@ -0,0 +1,153 @@
...
+                </div>
+                <hr>
+            </div>
+
+            <!--step 2-->
+            <div class="row" ng-show="syncItemStep == 2">
+                <div class="row" style="margin-top: 10px;">
+                    <div class="form-horizontal">
+                        <div class="col-sm-12">
+                            <label class="col-sm-2 control-label">是否比较注释</label>
```
nobodyiam: This style looks a little strange, as the screen shot below

image

With the following change, it may look a little better..

```html
                        <div class="col-sm-12">
                            <label class="control-label">
                                <input type="checkbox" class="comment-toggle" ng-checked="showCommentDiff"
                                       ng-click="showCommentDiff=!showCommentDiff">
                                是否比较注释
                            </label>
                        </div>
```
image

kezhenxu94: Thanks, updated the style

PR #2008, opened 2019-02-26
add 'exceptionReturnAspect.java',it will print infomation log when an exception occurs to an HTTP request。

jiangcaijun:
###  Scenario:
Multiple clients access the config service. When problems arise, it is impossible to locate which client and what requests were made according to the log.

###  Solution:
add 'exceptionReturnAspect.java',it will print log when an exception occurs to an HTTP request。

Output information includes IP、error_msg、url、request_param ...

###  Example:
HTTP eg:
> http://127.0.0.1:8080/notifications/v2?appId=11&cluster=default&notifications=

log out:
> 2019-02-26 16:53:30.344 ERROR 82278 --- [nio-8080-exec-3] c.c.f.a.c.aop.ExceptionReturnAspect      : {"error_msg":"Invalid format of notifications: 。other info:appId=11,cluster=default,dataCenter=null,clientIp=null","ip":"127.0.0.1","request_param":{"cluster":"default","appId":"11"},"url":"/notifications/v2"}


nobodyiam:
Thanks, but there is already a [GlobalDefaultExceptionHandler](https://github.com/ctripcorp/apollo/blob/master/apollo-common/src/main/java/com/ctrip/framework/apollo/common/controller/GlobalDefaultExceptionHandler.java) which has the similar logic, can we reuse it?

Also, I don't suggest including a new json dependency, can gson serve the purpose?

jiangcaijun:
thank you。
This is my first time to submit PR. I'm not sure about the open source community code submission model.I'll check it again and submit PR.

Thank you for your reply

PR #2014, opened 2019-02-28
add other info to printLog when an exception occurs to an HTTP reques…

jiangcaijun:
### Scenario:
Multiple clients access the config service. When problems arise, it is impossible to locate which client and what requests were made according to the log.

### Solution:
edit 'GlobalDefaultExceptionHandler.java',it will add additional information to log when an exception occurs to an HTTP request。The additional information includes ip、url、httpInputParams ...

### Example:
HTTP eg:

> curl http://192.168.0.1:8080/notifications/v2?appId=11&cluster=default&notifications=

log out:

> 2019-02-28 17:28:45.858 ERROR 99598 --- [nio-8080-exec-1] .c.f.a.c.c.GlobalDefaultExceptionHandler : message:Invalid format of notifications: ;url:http://192.168.0.1:8080/notifications/v2,httpInputParams:{cluster=default, appId=11},ip:192.168.40.34

### DIFF
--- a/apollo-common/src/main/java/com/ctrip/framework/apollo/common/controller/GlobalDefaultExceptionHandler.java
+++ b/apollo-common/src/main/java/com/ctrip/framework/apollo/common/controller/GlobalDefaultExceptionHandler.java
@@ -8,6 +8,7 @@ import com.google.gson.reflect.TypeToken;
 import java.lang.reflect.Type;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
+import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
@@ -75,7 +76,7 @@ public class GlobalDefaultExceptionHandler {
 
   @ExceptionHandler(MethodArgumentNotValidException.class)
   public ResponseEntity<Map<String, Object>> handleMethodArgumentNotValidException(
-      HttpServletRequest request, MethodArgumentNotValidException ex
+          HttpServletRequest request, MethodArgumentNotValidException ex
   ) {
     final Optional<ObjectError> firstError = ex.getBindingResult().getAllErrors().stream().findFirst();
     if (firstError.isPresent()) {
@@ -87,7 +88,7 @@ public class GlobalDefaultExceptionHandler {
 
   @ExceptionHandler(ConstraintViolationException.class)
   public ResponseEntity<Map<String, Object>> handleConstraintViolationException(
-      HttpServletRequest request, ConstraintViolationException ex
+          HttpServletRequest request, ConstraintViolationException ex
   ) {
     return handleError(request, BAD_REQUEST, new BadRequestException(ex.getMessage()));
   }
@@ -100,7 +101,7 @@ public class GlobalDefaultExceptionHandler {
   private ResponseEntity<Map<String, Object>> handleError(HttpServletRequest request,
                                                           HttpStatus status, Throwable ex, Level logLevel) {
     String message = ex.getMessage();
-    printLog(message, ex, logLevel);
+    printLog(message, ex, logLevel, request);
 
     Map<String, Object> errorAttributes = new HashMap<>();
     boolean errorHandled = false;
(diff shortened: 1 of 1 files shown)


apollo-common/src/main/java/com/ctrip/framework/apollo/common/controller/GlobalDefaultExceptionHandler.java
```diff
@@ -153,4 +156,44 @@ private void printLog(String message, Throwable ex, Level logLevel) {
...
+  }
+
+  /**
+   * 获取 ip 地址
+   * @param request
+   * @return
+   */
+  public String getIpAddr(HttpServletRequest request) {
+    String ip = request.getHeader("x-forwarded-for");
+    if (ip == null || ip.length() == 0 || "unknown".equalsIgnoreCase(ip)) {
```
nobodyiam: why would ip equals to "unknown?"

apollo-common/src/main/java/com/ctrip/framework/apollo/common/controller/GlobalDefaultExceptionHandler.java
```diff
@@ -153,4 +156,44 @@ private void printLog(String message, Throwable ex, Level logLevel) {
...
+   * @param request
+   * @return
+   */
+  private Map getInputParams(HttpServletRequest request) {
+    Map<String, String> map = new HashMap();
+    Enumeration paramNames = request.getParameterNames();
+    while (paramNames.hasMoreElements()) {
+      String paramName = (String) paramNames.nextElement();
+      String[] paramValues = request.getParameterValues(paramName);
+      if (paramValues.length == 1) {
```
nobodyiam: how about paramValues.length > 1?

apollo-common/src/main/java/com/ctrip/framework/apollo/common/controller/GlobalDefaultExceptionHandler.java
jiangcaijun: If the service is represented by some agents such as **Squid**，and The **forwarded_for** attribute is set to **off**，then **X-Forwarded-For** equals “unknown”

Reference link：https://www.cnblogs.com/softidea/p/5705297.html

apollo-common/src/main/java/com/ctrip/framework/apollo/common/controller/GlobalDefaultExceptionHandler.java
jiangcaijun: if:   **paramValues < 1**

in this case, the paramValue is meaningless, for a practical application scenario like Apollo。At the same time, the simplicity of post-event log audit is considered.

apollo-common/src/main/java/com/ctrip/framework/apollo/common/controller/GlobalDefaultExceptionHandler.java
```diff
@@ -131,7 +132,9 @@
   }
 
   //打印日志, 其中logLevel为日志级别: ERROR/WARN/DEBUG/INFO/TRACE
-  private void printLog(String message, Throwable ex, Level logLevel) {
+  private void printLog(String message, Throwable ex, Level logLevel, HttpServletRequest request) {
+    message = String.format("message:%s;url:%s,httpInputParams:%s,ip:%s",
```
nobodyiam: how about `"message:%s, url:%s, params:%s, ip:%s"`?

apollo-common/src/main/java/com/ctrip/framework/apollo/common/controller/GlobalDefaultExceptionHandler.java
nobodyiam: I think http supports one param with multiple values, such as `http://xxx.com?q=a&q=b`, though it's not a typical use case for apollo.

Since we want to expose detailed information in logs, I think it's better to log everything we have, e.g. for the above  case, log `q=a,b`.

Another minor point is ServletRequest has a method `getParameterMap` which may be easier to use since we are printing all the params.

apollo-common/src/main/java/com/ctrip/framework/apollo/common/controller/GlobalDefaultExceptionHandler.java
nobodyiam: Thanks for the knowledge sharing, actually apollo already has similar logic in `com.ctrip.framework.apollo.configservice.controller.ConfigController#tryToGetClientIp` which only considers the standard case, and I think maybe it's enough.  Because it's hard to enumerate all the proxy headers in the world...

PR #2017, opened 2019-03-03
[APOLLO-2015]: fix bug that cannot resolve blank line in properties under Windows

### LINKED ISSUE #2015: windows操作系统下复制粘贴配置内容时，如果有空行，会报错
如题，我的复制粘贴内容如下
```
spring.datasource.url = jdbc:mysql://172.16.113.135:3306/sandbox2?characterEncoding=utf8
url = https://${var.env}.xxx.com
var.env = dev
k2 = dev.v2
k1 = dev.v1

spring.datasource.driver-class-name = com.mysql.jdbc.Driver
spring.datasource.username = root
spring.datasource.password = root
raw = dev.raw.value
```

f40cf5a3-287c-449a-8ca9-d1624bff8774

kezhenxu94:
resolve #2015


apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/component/txtresolver/PropertyResolver.java
```diff
@@ -155,7 +155,7 @@ private boolean isBlankItem(ItemDTO item) {
   }
 
   private boolean isBlankItem(String line) {
-    return "".equals(line);
+    return  "".equals(line.trim());
```
nobodyiam: how about `Strings.nullToEmpty(line).trim().isEmpty()` which works even `line` is null?

kezhenxu94: I've thought about this, but seems `line` cannot be null in this case, anyway, it's a coding defection here, I agree with you

PR #2047, opened 2019-03-13
Enhance: Dynamically add namespace

yangdihangN:
在运行时监听新的配置表，新监听配置表的配置可根据优先级更新至SpringValue中


nobodyiam:
Is there a typical case to add namespace dynamically to Spring?

yangdihangN:
This feature can be used to degrade the service when the traffic is too large.
dynamicNamespaceForCircuitBreaker

nobodyiam:
Thanks for the explanation, is it a case to dynamically add namespace to override normal configurations?

However, for the case you mentioned(circuit breaker), why don't we use a flag to switch the logic? I think switching the logic might be easier to understand.

yangdihangN:
If we use hystrix, we can use apollo to delivery HystrixCommand. In the normal namespace, we can use a relatively loose fuse strategy. Otherwise, we can use a strict fuse strategy. 
This feature can change the configuration of an instance at runtime without affecting other instances., I think it makes apollo more flexible and safe in use。😄

PR #2049, opened 2019-03-13
enable [ EnableApolloConfig ]、[ ApolloConfigChangeListener ] get namespace value from configuration file

liubake:
enable [ EnableApolloConfig ]、[ ApolloConfigChangeListener ] get value from configuration file, the order is:

1、annotation value
2、configuration value
3、defalut value [ ConfigConsts.NAMESPACE_APPLICATION ]


nobodyiam:
Maybe we could achieve this goal by simply supporting spring placeholder in `EnableApolloConfig` and `ApolloConfigChangeListener`, for example:

`@EnableApolloConfig("${xxx}")`

and in spring property files like `application.properties`, the content could be

```properties
xxx=someNamespace
```

PR #2066, opened 2019-03-21
fix conflict logical in properties namesapce.

JaredTan95:
Verify whether to repeat in text mode, when adding `properties` config.

### DIFF
--- a/apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/component/txtresolver/PropertyResolver.java
+++ b/apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/component/txtresolver/PropertyResolver.java
@@ -81,7 +81,7 @@ public class PropertyResolver implements ConfigTextResolver {
         keyCount++;
         String[] kv = parseKeyValueFromItem(item);
         if (kv != null) {
-          keys.add(kv[0].toUpperCase());
+          keys.add(kv[0].toLowerCase());
         } else {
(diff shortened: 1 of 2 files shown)


apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/component/txtresolver/PropertyResolver.java
```diff
@@ -89,6 +89,12 @@ private boolean isHasRepeatKey(String[] newItems) {
       lineCounter++;
     }
 
+    keys.forEach(item -> {
```
nobodyiam: how about changing line 84 to `keys.add(kv[0].toLowerCase());`?

JaredTan95: In my opinion, this way would be more flexible. It won't limit the user to use all lowercase key in properties. :-P

nobodyiam: by changing the key to lowercase in line 84 won't affect the actual properties, it's only a way to check whether there are case-insensitive duplicated keys...

JaredTan95: Aha, I got this point. :-D

PR #2067, opened 2019-03-21
add APP_ID env.

JaredTan95:


apollo-core/src/main/java/com/ctrip/framework/foundation/internals/provider/DefaultApplicationProvider.java
```diff
@@ -86,7 +86,15 @@ private void initAppId() {
       return;
     }
 
-    // 2. Try to get app id from app.properties.
+    //2. Try to get app id from OS environment variable
+    m_appId = System.getenv("APOLLO_APP_ID");
```
nobodyiam: why not just `APP_ID`?

JaredTan95: I just think `APOLLO_APP_ID ` won't conflict with user's own env variables. `APP_ID` still keep with `app.id`, this would be more friendly. OK, I will change to `APP_ID`.

apollo-core/src/main/java/com/ctrip/framework/foundation/internals/provider/DefaultApplicationProvider.java
```diff
@@ -86,7 +86,15 @@ private void initAppId() {
       return;
     }
 
-    // 2. Try to get app id from app.properties.
+    //2. Try to get app id from OS environment variable
+    m_appId = System.getenv("APP_ID");
+    if (!Utils.isBlank(m_appId)) {
+      m_appId = m_appId.trim();
+      logger.info("App ID is set to {} by APP_ID env from OS environment variable", m_appId);
```
nobodyiam: one minor typo fix:

`App ID is set to {} by APP_ID property from OS environment variable`

PR #2224, opened 2019-05-13
Update application.yml

soxueren:


soxueren:
cloud:
        inetutils:
          ignoredInterfaces:
            - docker0
            - veth.*

nobodyiam:
How about adding it as a -D parameter in JAVA_OPTS in startup.sh so that people could change it easily? I'm not sure whether it's applicable to every corner cases...

PR #2349, opened 2019-06-18
spring-boot-starter -> spring-boot-starter-test

slievrly:
dependency:

spring-boot-starter -> spring-boot-starter-test

### DIFF
--- a/apollo-mockserver/pom.xml
+++ b/apollo-mockserver/pom.xml
@@ -47,7 +47,7 @@
     </dependency>
     <dependency>
       <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-starter</artifactId>
+      <artifactId>spring-boot-starter-test</artifactId>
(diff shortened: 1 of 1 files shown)


slievrly:
ConfigControllerIntegrationTest.testQueryPublicConfigWithDataCenterFoundAndNoOverride:234 » HttpClientError

apollo-mockserver/pom.xml
```diff
@@ -47,7 +47,7 @@
     </dependency>
     <dependency>
       <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-starter</artifactId>
+      <artifactId>spring-boot-starter-test</artifactId>
```
nobodyiam: looks like it's not necessary as `spring-boot-starter-test` is included as a test scope dependency in [the parent pom](https://github.com/ctripcorp/apollo/blob/master/pom.xml#L354)

PR #2507, opened 2019-08-08
add prometheus integration.

JaredTan95:
support https://github.com/ctripcorp/apollo/issues/2489

- Exposed prometheus API: `/prometheus`  in Apollo-ConfigService, Apollo-Adminservice,Apollo-Portal modules.
  e.g.
  - localhost:8080/prometheus
  - localhost:8090/prometheus
  - localhost:8070/prometheus

- How to use: https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-metrics-export-prometheus

### DIFF
--- a/apollo-common/pom.xml
+++ b/apollo-common/pom.xml
@@ -61,12 +61,7 @@
             <artifactId>commons-lang</artifactId>
         </dependency>
 
-        <!-- Spring boot actuator to expose metrics endpoint -->
-        <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-actuator</artifactId>
-        </dependency>
-        <!-- Micormeter core dependecy -->
+        <!-- Micrometer core dependecy -->
         <dependency>
             <groupId>io.micrometer</groupId>
             <artifactId>micrometer-core</artifactId>
--- a/apollo-common/pom.xml
+++ b/apollo-common/pom.xml
@@ -60,5 +60,20 @@
             <groupId>commons-lang</groupId>
             <artifactId>commons-lang</artifactId>
         </dependency>
+
+        <!-- Spring boot actuator to expose metrics endpoint -->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-actuator</artifactId>
+        </dependency>
+        <!-- Micormeter core dependecy -->
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-registry-prometheus</artifactId>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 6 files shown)


apollo-common/src/main/java/com/ctrip/framework/apollo/common/controller/WebMvcConfig.java
```diff
@@ -29,7 +27,7 @@ public void addArgumentResolvers(List<HandlerMethodArgumentResolver> argumentRes
   @Override
   public void configureContentNegotiation(ContentNegotiationConfigurer configurer) {
     configurer.favorPathExtension(false);
-    configurer.ignoreAcceptHeader(true).defaultContentType(MediaType.APPLICATION_JSON_UTF8);
+    //configurer.ignoreAcceptHeader(true).defaultContentType(MediaType.APPLICATION_JSON_UTF8);
```
nobodyiam: It's better to remove it if it's not necessary.

apollo-common/src/main/resources/application.properties
```diff
@@ -17,3 +17,5 @@ server.max-http-header-size=10240
 management.endpoints.web.base-path=/
 logging.file.max-size=50MB
 logging.file.max-history=10
+
+management.endpoints.web.exposure.include=*
```
nobodyiam: It is dangerous to expose all the endpoints, the default is `info, health`, so how about change it to `info,health, prometheus `

pom.xml
```diff
@@ -362,6 +362,21 @@
 				</exclusion>
 			</exclusions>
 		</dependency>
+
+		<!-- Spring boot actuator to expose metrics endpoint -->
+		<dependency>
```
nobodyiam: These dependencies are only needed in config-service/admin-service and portal, so they should be moved to `apollo-common`.

Besides, I think they also caused the travis-ci build failure for apollo-core module.

apollo-common/pom.xml
```diff
@@ -60,5 +60,20 @@
             <groupId>commons-lang</groupId>
             <artifactId>commons-lang</artifactId>
         </dependency>
+
+        <!-- Spring boot actuator to expose metrics endpoint -->
+        <dependency>
```
nobodyiam: it looks like `spring-boot-starter-actuator` was already delcared in line 22...

apollo-common/pom.xml
```diff
@@ -60,5 +60,20 @@
...
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-actuator</artifactId>
+        </dependency>
+        <!-- Micormeter core dependecy -->
```
nobodyiam: Micormeter => Micrometer ?

apollo-common/src/main/resources/application.properties
```diff
@@ -17,3 +17,7 @@ server.max-http-header-size=10240
 management.endpoints.web.base-path=/
 logging.file.max-size=50MB
 logging.file.max-history=10
+
+management.endpoints.web.exposure.include=info,health,metrics,prometheus
+management.endpoint.metrics.enabled=true
```
nobodyiam: According to [Enabling Endpoints](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html#production-ready-endpoints-enabling-endpoints): by default, all endpoints except for shutdown are enabled. So it's not necessary to set metrics and prometheus endpoints enabled?

JaredTan95: No need indeed, just to show other developers. Or delete it.

apollo-common/pom.xml
JaredTan95: yeah, typo

PR #2543, opened 2019-08-24
rename timeout to timeoutInMilli for better understanding

nobodyiam:


kezhenxu94:
Seems `millis` is more commonly-used as `millisecond` in Java, such as `java.lang.System#currentTimeMillis` and `java.util.concurrent.TimeUnit#toMillis`, so `timeoutInMillis` may be familiar to Java developer

nobodyiam:
@kezhenxu94 

Thanks for you suggestion, and I now think `millis` is better than `milli`.
However, there are many places in the project using `milli` instead of `millis`, so how about we keep the consistency for now?

kezhenxu94:
> @kezhenxu94
> 
> Thanks for you suggestion, and I now think `millis` is better than `milli`.
> However, there are many places in the project using `milli` instead of `millis`, so how about we keep the consistency for now?

Sure, consistency is also important, then merge it :)

PR #2646, opened 2019-10-10
fix setting null value of property in xml throws NullPointerException

xuanuu:
解决如下问题：
当propertyString==null时，会抛空指针异常NullPointerException。例如：在xml的配置中，如果设置属性值为null， `<property name="manager"></null></property>`，即发生propertyString==null

### DIFF
--- a/apollo-client/src/main/java/com/ctrip/framework/apollo/spring/property/PlaceholderHelper.java
+++ b/apollo-client/src/main/java/com/ctrip/framework/apollo/spring/property/PlaceholderHelper.java
@@ -4,6 +4,7 @@ import com.google.common.base.Strings;
 import com.google.common.collect.Sets;
 import java.util.Set;
 import java.util.Stack;
+import java.util.Objects;
 import org.springframework.beans.factory.BeanFactory;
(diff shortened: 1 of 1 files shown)


apollo-client/src/main/java/com/ctrip/framework/apollo/spring/property/PlaceholderHelper.java
```diff
@@ -65,7 +65,7 @@ private Object evaluateBeanDefinitionString(ConfigurableBeanFactory beanFactory,
   public Set<String> extractPlaceholderKeys(String propertyString) {
     Set<String> placeholderKeys = Sets.newHashSet();
 
-    if (!isNormalizedPlaceholder(propertyString) && !isExpressionWithPlaceholder(propertyString)) {
+    if (propertyString == null || (!isNormalizedPlaceholder(propertyString) && !isExpressionWithPlaceholder(propertyString))) {
```
JaredTan95: LGTM
BTW, How about use `java.util.Objects.isNull` here.

nobodyiam:
Looks good! It will be perfect if some unit tests are added to cover this scenario.

PR #2678, opened 2019-10-19
Bugfix: fix miscellaneous bugs and polish code

kezhenxu94:
This patch fixes miscellaneous bugs, such as `{}` in `String.format`, possible NPE.
Some performance optimization, such as unnecessary primitive boxing and unboxing, inner class can be `static`

### DIFF
--- a/apollo-mockserver/src/main/java/com/ctrip/framework/apollo/mockserver/EmbeddedApollo.java
+++ b/apollo-mockserver/src/main/java/com/ctrip/framework/apollo/mockserver/EmbeddedApollo.java
@@ -67,6 +67,8 @@ public class EmbeddedApollo extends ExternalResource {
         } else if (request.getPath().startsWith("/configs")) {
           List<String> pathSegments = request.getRequestUrl().pathSegments();
           // appId and cluster might be used in the future
+          String appId = pathSegments.get(1);
+          String cluster = pathSegments.get(2);
           String namespace = pathSegments.get(3);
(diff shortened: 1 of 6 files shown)


apollo-mockserver/src/main/java/com/ctrip/framework/apollo/mockserver/EmbeddedApollo.java
```diff
@@ -68,8 +67,6 @@ public MockResponse dispatch(RecordedRequest request) throws InterruptedExceptio
         } else if (request.getPath().startsWith("/configs")) {
           List<String> pathSegments = request.getRequestUrl().pathSegments();
           // appId and cluster might be used in the future
-          String appId = pathSegments.get(1);
```
JaredTan95: how about keeping it?

nobodyiam: I would also prefer keeping it, as it might be hard future maintainers to understand how to retrieve the appId and cluster

kezhenxu94:   I missed the comment above the codes

kezhenxu94: Reverted back

PR #2724, opened 2019-11-03
support search by item key

lepdou:
### 1. 支持按配置项的key搜索
image


apollo-biz/src/main/java/com/ctrip/framework/apollo/biz/repository/ItemRepository.java
```diff
@@ -19,6 +19,8 @@
 
   List<Item> findByNamespaceIdAndDataChangeLastModifiedTimeGreaterThan(Long namespaceId, Date date);
 
+  List<Item> findByKeyLike(String key);
```
nobodyiam: There is no index on column key, which will be very slow.

lepdou: 这里现在 key 没有拼上 %，所以现在没有走 like 操作。
可以给 key 加个索引，走索引的情况下，百万级别数据量应该都没啥问题？

lepdou:
> This is a great feature. However, since we don't involve any search engine, it is kind of slow to search from database, especially when there are many items in the db.
> 
> Will it be better if we provide a flag to let the user decide whether to enable this feature or not?
> Or is it better to provide a new entry to search items?

可以先简单点，搞个配置项，默认开着，因为大部分用户都达不到这个量级。
等有用户反馈搜索太慢，可以再优化。

apollo-biz/src/main/java/com/ctrip/framework/apollo/biz/repository/ItemRepository.java
nobodyiam: sql语句里面是Like，加不加%都会走全表扫描的，所以索引还是得加下。

image

加了索引之后

image

nobodyiam:
> > This is a great feature. However, since we don't involve any search engine, it is kind of slow to search from database, especially when there are many items in the db.
> > Will it be better if we provide a flag to let the user decide whether to enable this feature or not?
> > Or is it better to provide a new entry to search items?
> 
> 可以先简单点，搞个配置项，默认开着，因为大部分用户都达不到这个量级。
> 等有用户反馈搜索太慢，可以再优化。

默认是不是关掉比较好？因为可能不是所有的用户都需要这个功能，而且这个可能会对数据库带来比较大的压力

hezhaoye:
请教下，这个功能为啥没有merge到master分支呢

PR #2826, opened 2019-12-07
Dynamic env

Anilople:
Dynamic environment setting when portal service start.
A new way to forbid changing hard code to add new environment.
Just change 2 configuration properties.
Suppose that you want to add a new environment beta, and you have been add dev, just do like follow step.
1. apollo.portal.envs=dev,beta
2. beta.meta=your-meta-address


Anilople:
a new PR has been commit

PR #2828, opened 2019-12-10
Implement  #2099，添加apollo-client的身份认证功能及accesskey管理功能

nisiyong:

- 添加apollo-configservice对apollo-client的身份认证
- 添加apollo-portal的accesskey管理功能

Implement  #2099

### apollo-client

1. 在客户端添加配置`apollo.accesskey.secret`，用来做访问apollo对密钥
2. 用secret对请求签名，在客户端访问configservice对请求添加请求头`Authorization`,`Timestamp`

### apollo-configservice

1. 添加filter过滤器，解析header，对客户端请求进行身份认证，如果非法请求则返回401
2. 对于旧版项目没有配置secret对，默认跳过。等添加了secret才会进行身份认证
3. `ApolloConfigDB`添加新表`AccessKey`
```sql
CREATE TABLE `AccessKey` (
  `Id` int(10) unsigned NOT NULL AUTO_INCREMENT COMMENT '自增主键',
  `AppId` varchar(500) NOT NULL DEFAULT 'default' COMMENT 'AppID',
  `Secret` varchar(128) NOT NULL DEFAULT '' COMMENT 'Secret',
  `IsEnabled` bit(1) NOT NULL DEFAULT b'0' COMMENT '1: enabled, 0: disabled',
  `IsDeleted` bit(1) NOT NULL DEFAULT b'0' COMMENT '1: deleted, 0: normal',
  `DataChange_CreatedBy` varchar(32) NOT NULL DEFAULT 'default' COMMENT '创建人邮箱前缀',
  `DataChange_CreatedTime` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT '创建时间',
  `DataChange_LastModifiedBy` varchar(32) NOT NULL DEFAULT '' COMMENT '最后修改人邮箱前缀',
  `DataChange_LastTime` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT '最后修改时间',
  PRIMARY KEY (`Id`),
  KEY `AppId` (`AppId`(191)),
  KEY `DataChange_LastTime` (`DataChange_LastTime`)
) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8mb4 COMMENT='访问密钥';
```

### apollo-portal & apollo-adminservice

> 感谢 @huangchenhai-666 提供Portal页面支持
1. 添加`密钥管理`功能
2. 支持多个密钥（最多5个），用于密钥更替时兼容
3. 已创建的项目需要添加权限。可以用`insert select`清洗数据
```sql
INSERT INTO `Permission` (`PermissionType`, `TargetId`, `IsDeleted`, `DataChange_CreatedBy`, `DataChange_LastModifiedBy`) VALUES ( 'ManageAppMaster', 'demo', 0, 'apollo', 'apollo');
```

## Featrue Preview

项目管理员可以看到`管理密钥`对菜单
image

为每个项目配置密钥，每个环境密钥不同。选择环境添加后自动生成，默认是禁用的
image


apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/component/PermissionValidator.java
```diff
@@ -102,6 +102,12 @@ public boolean hasCreateClusterPermission(String appId) {
         appId);
   }
 
+  public boolean hasManageAccessKeyPermission(String appId) {
```
nobodyiam: how about just use `isAppAdmin` so that we don't need to a new permission type?

apollo-biz/src/main/java/com/ctrip/framework/apollo/biz/service/AccessKeyService.java
```diff
@@ -0,0 +1,83 @@
...
+      AuditService auditService) {
+    this.accessKeyRepository = accessKeyRepository;
+    this.auditService = auditService;
+  }
+
+  public List<AccessKey> findByAppId(String appId) {
+    return accessKeyRepository.findByAppId(appId);
+  }
+
+  @Transactional(rollbackFor = Exception.class)
```
nobodyiam: Is it really necessary to specify `rollbackFor` here? As other parts of apollo codes don't have this and use default rollbackFor.

apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/filter/AccessFilter.java
```diff
@@ -0,0 +1,116 @@
...
+      // check signature
+      String path = request.getServletPath();
+      String query = request.getQueryString();
+      if (!checkAuthorization(authorization, availableSecrets, timestamp, path, query)) {
+        logger.warn("Invalid authorization. appId={},authorization={}", appId, authorization);
+        response.sendError(HttpServletResponse.SC_UNAUTHORIZED, "Unauthorized");
+        return;
+      }
+
+      chain.doFilter(request, response);
```
nobodyiam: I think we could simplify this by removing the else statment and call `chain.doFilter(request, response);` directly.

```java
    if (!CollectionUtils.isEmpty(availableSecrets)) {
      String timestamp = request.getHeader(Signature.HTTP_HEADER_TIMESTAMP);
      String authorization = request.getHeader(Signature.HTTP_HEADER_AUTHORIZATION);
      ...
    }
    chain.doFilter(request, response);
```

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/component/PermissionValidator.java
nisiyong: Yeah, `isAppAdmin` is what I want! 
I'll will polish this logic code.

nisiyong: `com.ctrip.framework.apollo.portal.component.PermissionValidator#isAppAdmin`
```java
public boolean isAppAdmin(String appId) {
  return isSuperAdmin() || hasAssignRolePermission(appId);
}
```
But I found `isAppAdmin` means whether it has `AssignRole` permission. We should aware that `AssignRole` permission has two meanings after we use it for `ManageAccessKey` permission.

apollo-biz/src/main/java/com/ctrip/framework/apollo/biz/service/AccessKeyService.java
nisiyong: It doesn't matter to specify`rollbackFor`, I add it cause I have installed `Alibaba Java Coding Guidelines` IDEA plugin, and the plugin advises me to add it.
image
I can remove all `rollbackFor` to keep it as same as other codes.

apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/filter/AccessFilter.java
nisiyong: Only removing the else statment will change the logic, the `chain.doFilter(request, response)` will execute 2 times.
```java
if (!CollectionUtils.isEmpty(availableSecrets)) {      
    ...
    chain.doFilter(request, response);
}
chain.doFilter(request, response);
```
 There have two choices to do that:
First one, With `return` statement
```java
if (!CollectionUtils.isEmpty(availableSecrets)) {      
    ...
    chain.doFilter(request, response);
    return;
}
chain.doFilter(request, response);
```
Another one, without `return` which I had wrote.
```java
if (!CollectionUtils.isEmpty(availableSecrets)) {      
    ...
    chain.doFilter(request, response);
} else {
    chain.doFilter(request, response);
}
```

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/component/PermissionValidator.java
nobodyiam: > But I found isAppAdmin means whether it has AssignRole permission. We should aware that AssignRole permission has two meanings after we use it for ManageAccessKey permission.

That should be OK.

nisiyong:
@nobodyiam Hi, Song. Thank you for your review and for some advice. I have resolved some problems that you mentioned above, and this PR has some changes as follows:
1. Use `isAppAdmin` replace `hasManageAccessKeyPermission`, then we just need to create `AccessKey` table  with the upgrade, no more DML to fix historical data.
2. Add icon and some tips for AccessKey Management in the portal, it is more beautiful than before.
3. Rename`AccessFilter` to`ClientAuthenticationFilter`.
4. Add some unit tests for core modules.

Please take a look.

## Preview
image
image
image

apollo-biz/src/test/java/com/ctrip/framework/apollo/biz/repository/AccessKeyRepositoryTest.java
```diff
@@ -0,0 +1,50 @@
...
+    accessKey.setAppId(appId);
+    accessKey.setSecret(secret);
+    accessKey.setEnabled(true);
+
+    assertEquals(0, accessKeyRepository.count());
+    accessKeyRepository.save(accessKey);
+    assertEquals(1, accessKeyRepository.count());
+
+    List<AccessKey> accessKeyList = accessKeyRepository.findByAppId(appId);
+    assertEquals(1, accessKeyList.size());
```
nobodyiam: It is better to add more assert statements to test the field values, such as appId and secret?

apollo-client/src/test/java/com/ctrip/framework/apollo/internals/RemoteConfigLongPollServiceTest.java
```diff
@@ -476,6 +475,11 @@ public String getCluster() {
       return someCluster;
     }
 
+    @Override
+    public String getAccessKeySecret() {
+      return someSecret;
```
nobodyiam: shall we add some tests to test whether secret is in the http request header?

And I think it's better to add different scenarios to test when the secret is disabled, which is the default behavior.

apollo-client/src/test/java/com/ctrip/framework/apollo/internals/RemoteConfigRepositoryTest.java
```diff
@@ -262,6 +259,11 @@ public String getCluster() {
       return "someCluster";
     }
 
+    @Override
+    public String getAccessKeySecret() {
+      return "someSecret";
```
nobodyiam: the same as above.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/spi/defaultimpl/DefaultRoleInitializationService.java
```diff
@@ -15,14 +15,13 @@
...
 import com.google.common.collect.Sets;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.transaction.annotation.Transactional;
-
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.springframework.beans.factory.annotation.Autowired;
```
nobodyiam: I think we could revert the changes of this file as it has nothing actually changed.

apollo-biz/src/test/java/com/ctrip/framework/apollo/biz/repository/AccessKeyRepositoryTest.java
nisiyong: OK

nisiyong:
@nobodyiam I have fixed the problems above, the changes are as follow:
1. Revert some files that have nothing changes.
2. Add new unit test case about AccessKeySecret in `RemoteConfigRepositoryTest` and `RemoteConfigLongPollServiceTest`
3. Squash git commits. There are 2 commits in this PR.

nobodyiam:
@nisiyong 

I have tested in my local environment and it looked great!

However, it seems it will take up to 1 minute to take effect after I enable/disable the access key. It may be a problem in some situations. 

Can we change the implementation of `com.ctrip.framework.apollo.configservice.service.AccessKeyServiceWithCache#scanNewAccessKeys` to make it faster?

nisiyong:
> However, it seems it will take up to 1 minute to take effect after I enable/disable the access key. It may be a problem in some situations.

Yes, the cache is refreshed every minute by default. In my opinion, I don't think it is a problem because we don't change the secret often. Maybe you can tell me more details about this in some situations.



> Can we change the implementation of `com.ctrip.framework.apollo.configservice.service.AccessKeyServiceWithCache#scanNewAccessKeys` to make it faster?

Do you mean to change the default `rebuildInterval`? I think I can set it 10 seconds by default.

nobodyiam:
@nisiyong 

Suppose the user enable the secret and just to find the applications printing error logs (because of no secret), so he decides to disable the secret temporarily. In this situation, he has to wait for up to 1 minute to see the effect, which I think will confuse the user.

What I suggest is to change the implementation, e.g. find the new/updated access key by DataChange_LastTime, I think this can solve the problem.

nisiyong:
> What I suggest is to change the implementation, e.g. find the new/updated access key by DataChange_LastTime, I think this can solve the problem.

@nobodyiam 

I have considered this implementation before, but the deletion of the access key will still take effect one minute by default. It is indeed better to use `DataChange_LastTime` to find the new/updated access key every second by default. 

Okay, I will change the implementation this week.

apollo-biz/src/test/java/com/ctrip/framework/apollo/biz/repository/AccessKeyRepositoryTest.java
```diff
@@ -0,0 +1,66 @@
...
+    assertThat(accessKeyList.get(0).getAppId()).isEqualTo(appId);
+    assertThat(accessKeyList.get(0).getSecret()).isEqualTo("someSecret");
+  }
+
+  @Test
+  @Sql(scripts = "/sql/accesskey-test.sql", executionPhase = ExecutionPhase.BEFORE_TEST_METHOD)
+  @Sql(scripts = "/sql/clean.sql", executionPhase = ExecutionPhase.AFTER_TEST_METHOD)
+  public void testFindFirst500ByDataChangeLastModifiedTimeGreaterThanOrderByDataChangeLastModifiedTime() {
+    // 2019-12-19 10:46:35
+    Date date = new Date(1576723194000L);
```
nobodyiam: This causes the ci failed, because `1576723194000L` is 2019-12-19 10:39:54 in UTC+8, which is 2019-12-19 02:39:54 in UTC+0, so there will be 3 records returned instead of 2.

You may consider to use Calendar to create the date instance.

apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/service/AccessKeyServiceWithCache.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  private void loadNewAndUpdatedAccessKeys() {
+    boolean hasMore = true;
+    while (hasMore && !Thread.currentThread().isInterrupted()) {
+      //current batch is 500
+      Date fromTime = new Date(lastTimeScanned);
+      List<AccessKey> accessKeys = accessKeyRepository
+          .findFirst500ByDataChangeLastModifiedTimeGreaterThanOrderByDataChangeLastModifiedTimeAsc(fromTime);
```
nobodyiam: I see the current logic is `where DataChange_LastTime > xxx`, but I think we need to change it to `where DataChange_LastTime >= xxx` or we may miss some records.

e.g. there are 2 records with the same DataChange_LastTime, if the last scan only contains the first record, then the next records won't be scanned forever.

But with `where DataChange_LastTime >= xxx`, it will cause the last record scanned everytime, so maybe we could add more conditions like `where DataChange_LastTime >= xxx and id != yyy`

apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/service/AccessKeyServiceWithCache.java
```diff
@@ -0,0 +1,202 @@
...
+      if (CollectionUtils.isEmpty(accessKeys)) {
+        break;
+      }
+
+      mergeAccessKeys(accessKeys);
+
+      int scanned = accessKeys.size();
+      lastTimeScanned = accessKeys.get(scanned - 1).getDataChangeLastModifiedTime().getTime();
```
nobodyiam: why not just use the date object so that we don't need to create a new date for every scan?

apollo-portal/src/main/resources/static/app/access_key.html
```diff
@@ -0,0 +1,168 @@
...
+                                    <td style="text-align: center;">{{accessKey.dataChangeLastModifiedBy}}</td>
+                                    <td style="text-align: center;">{{accessKey.dataChangeLastModifiedTime | date: 'yyyy-MM-dd HH:mm:ss'}}</td>
+                                    <td style="text-align: center;">
+                                        <a href="javascript:;"
+                                           ng-click="enable(accessKey.id, env)" ng-if="!accessKey.enabled">{{'AccessKey.Operator.Enable' | translate}}</a>
+                                        <a href="javascript:;"
+                                           ng-click="disable(accessKey.id, env)" ng-if="accessKey.enabled">{{'AccessKey.Operator.Disable' | translate}}</a>
+                                        <a href="javascript:;"
+                                           ng-click="remove(accessKey.id, env)">{{'AccessKey.Operator.Remove' | translate }}</a>
```
nobodyiam: since the remove will take up to 1 minute to take effect, shall we only allow removing disabled secret keys? e.g. when the user wants to remove an enabled key, we promt him to disable the sceret first?

apollo-portal/src/main/resources/static/i18n/en.json
```diff
@@ -649,6 +679,7 @@
   "App.AppOwnerTips": "(After enabling the application administrator allocation restrictions, the application owner and project administrator are default to current account, not subject to change)",
   "App.AppAdminTips1": "(The application owner has project administrator permission by default.",
   "App.AppAdminTips2": "Project administrators can create namespace, cluster, and assign user permissions)",
+  "App.AccessKey.NoPermissionTips": "You do not have permission to operate, please ask [{{users}}]",
```
nobodyiam: You do not have permission to operate, please ask [{{users}}] to authorize

apollo-portal/src/main/resources/static/i18n/zh-CN.json
```diff
@@ -649,6 +679,7 @@
   "App.AppOwnerTips": "(开启项目管理员分配权限控制后，应用负责人和项目管理员默认为本账号，不可选择)",
   "App.AppAdminTips1": "(应用负责人默认具有项目管理员权限，",
   "App.AppAdminTips2": "项目管理员可以创建Namespace和集群、分配用户权限)",
+  "App.AccessKey.NoPermissionTips": "您没有权限操作，请找 [{{users}}]",
```
nobodyiam: 您没有权限操作，请找 [{{users}}] 开通权限

apollo-portal/src/main/resources/static/app/access_key.html
nisiyong: That's a good idea 👍!!!

apollo-biz/src/test/java/com/ctrip/framework/apollo/biz/repository/AccessKeyRepositoryTest.java
nisiyong: Thank you, and I think I can use `Instant` to create `Date`, like this:
```
Instant instant = LocalDateTime.of(2019, 12, 19, 10, 39, 54)
    .atZone(ZoneId.systemDefault())
    .toInstant();
Date date = Date.from(instant);
```

apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/service/AccessKeyServiceWithCache.java
nisiyong: Thank you for your careful review, this is indeed my mistake.

But if there are a lot of records with the same `DataChange_LastTime `, maybe we should use the condition like `where DataChange_LastTime >= xxx and id not in (xxx)`.

And suppose we use `limit 500`, if there are less than 500 records with the same `DataChange_LastTime`, the condition above will work fine. If more than 1000 records with the same `DataChange_LastTime`, the condition will cause an infinite loop.

nisiyong: How about I still use `where DataChange_LastTime > ? limit 500` and use `where DataChange_LastTime = ? limit 500` in each loop.

I think this is a simple way.

apollo-portal/src/main/resources/static/app/access_key.html
nisiyong: How about hiding the `Remove` button directly when it is enabled?

nisiyong: image

apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/service/AccessKeyServiceWithCache.java
nobodyiam: You are right about the infinite loop and I think your idea to check `where DataChange_LastTime = ? in each loop if the size of the results is equal to 500` is workable.

apollo-portal/src/main/resources/static/app/access_key.html
nobodyiam: This is a good idea.

apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/service/AccessKeyServiceWithCache.java
```diff
@@ -0,0 +1,206 @@
...
+      }
+
+      int scanned = accessKeys.size();
+      mergeAccessKeys(accessKeys);
+      logger.info("Loaded {} new/updated Accesskey from startTime {}", scanned, lastTimeScanned);
+
+      hasMore = scanned == 500;
+      lastTimeScanned = accessKeys.get(scanned - 1).getDataChangeLastModifiedTime();
+
+      if (hasMore) {
```
nobodyiam: It is better to add some comments here so that people will know why we need this logic here.

nisiyong: Okay

nisiyong:
@nobodyiam 

I have already solved all the problems you mentioned. If there are other problems, I will be happy to solve them.

PR #2836, opened 2019-12-12
客户端使用LinkedHashMap存储配置

### LINKED ISSUE #2451: 《使用apollo配置zuul动态路由不能保证数据有序性的问题》
我在apollo中的配置如下

Zuul启动后读取到的配置是无序的，如下

这样就会导致在请求 /user/debug/** 时，在做match()的时候被  /user/**  先匹配到了，然后返回并结束匹配过程

这样就会导致 /user/debug/** 匹配的服务始终不能被访问，只有 /user/** 匹配的服务能访问。
可能我描述的不是很清楚，另附上一篇 Zuul的路径匹配文章：
http://blog.didispace.com/spring-cloud-tips-zuul-path-config/

songdragon:

Keep order of configuration in yml/yaml at client side.

Fixes #2451, #2273

## Brief changelog

1. Use LinkedHashMap instead of Properties in apollo-client module.
2. Use LinkedHashMap to keep order in DefaultConfig.java Line 112.

### DIFF
--- a/apollo-client/src/main/java/com/ctrip/framework/apollo/spring/boot/ApolloApplicationContextInitializer.java
+++ b/apollo-client/src/main/java/com/ctrip/framework/apollo/spring/boot/ApolloApplicationContextInitializer.java
@@ -72,8 +72,7 @@ public class ApolloApplicationContextInitializer implements
   public void initialize(ConfigurableApplicationContext context) {
     ConfigurableEnvironment environment = context.getEnvironment();
 
-    String enabled = environment.getProperty(PropertySourcesConstants.APOLLO_BOOTSTRAP_ENABLED, "false");
-    if (!Boolean.valueOf(enabled)) {
+    if (!environment.getProperty(PropertySourcesConstants.APOLLO_BOOTSTRAP_ENABLED, Boolean.class, false)) {
       logger.debug("Apollo bootstrap config is not enabled for context {}, see property: ${{}}", context, PropertySourcesConstants.APOLLO_BOOTSTRAP_ENABLED);
       return;
     }


songdragon:
> Thanks for your contribution!
> 
> But it's not OK to change the public methods' signature directly, as it's not backward compatible. We need to put more effort on this.

 It seems changes affcet in apollo-client, "properties format" local cache may be loaded with exceptions. 

Is there any other scenarios should be taken into consideration? Or could you give more concerns on backward compatibility scenarios?

nobodyiam:
@songdragon 

People may already use these **public** methods in their code. If we change the signature, then these code will not work when they upgrade apollo-client version.

songdragon:
@nobodyiam 
You are right. So I am trying on following solutions:
 - [x] 1. Mark public methods with Properties as Deprecated
 - [x] 2. Add new public methods signature with Map, because both Properties and LinkedHashMap implmentes Map interface 
 - [x] 3, Move implmentation into new methods 
 - [x] 4. Delgate implmentations in 1 to new mthods (changed in 3)
 - [x] 5. fix tests with new methods mock and fix asserts condition

So far code looks work correctly, but there are two new problems:
 - [ ] 6. Methods with Property return value will return a new Object. This is different from previous version, and it causes some UT failure. 
I think it should return a cached Property if LinkedHashMap is a cached Object. Right?
 - [ ] 7. Add some test cases. I am not sure new implmentation will keep original order, and some tests should be add to validate this feature.

Please let me know if you have any idea. Cheers! :)

nobodyiam:
I think we introduced too many changes only to achieve this funcionality.

Is it possible we introduce an `OrderedProperties` which extends `Properties` but can keep the order?  If we can do so, then the changes will be much less.

I found a similar implementation: [OrderedProperties.java](https://github.com/etiennestuder/java-ordered-properties/blob/master/src/main/java/nu/studer/java/util/OrderedProperties.java), which we may reference as a start.

songdragon:
It should be work, but there may be some methods in Properties or Hastable should be override . Let's have a try.

songdragon:
@nobodyiam please review new PR: #2861 .

PR #2844, opened 2019-12-16
add feature for dynamic log level change under spring project

lawrencewu:

添加了一个能够在spring项目下动态更新日志级别的功能

Fixes #
https://github.com/ctripcorp/apollo/issues/2798
https://github.com/ctripcorp/apollo/issues/2482

## Brief changelog

增加了继承ConfigChangeListener的监听器
修改了<java.version>1.8</java.version>

### DIFF
--- a/apollo-client/pom.xml
+++ b/apollo-client/pom.xml
@@ -11,7 +11,7 @@
 	<artifactId>apollo-client</artifactId>
 	<name>Apollo Client</name>
 	<properties>
-		<java.version>1.7</java.version>
+		<java.version>1.8</java.version>
 		<github.path>${project.artifactId}</github.path>
 	</properties>
 	<dependencyManagement>
--- a/apollo-client/src/main/java/com/ctrip/framework/apollo/listener/LogLevelListener.java
+++ b/apollo-client/src/main/java/com/ctrip/framework/apollo/listener/LogLevelListener.java
@@ -1,0 +1,60 @@
+package com.ctrip.framework.apollo.listener;
+
+import com.ctrip.framework.apollo.ConfigChangeListener;
+import com.ctrip.framework.apollo.model.ConfigChange;
+import com.ctrip.framework.apollo.model.ConfigChangeEvent;
+import com.ctrip.framework.apollo.spring.annotation.ApolloConfigChangeListener;
+import org.springframework.boot.logging.LogLevel;
+import org.springframework.boot.logging.LoggingSystem;
+import org.springframework.stereotype.Service;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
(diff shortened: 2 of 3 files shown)


apollo-client/src/main/java/com/ctrip/framework/apollo/model/ConfigChangeEvent.java
```diff
@@ -55,4 +55,8 @@ public boolean isChanged(String key) {
   public String getNamespace() {
     return m_namespace;
   }
+
+  public Map<String, ConfigChange> getChangeValue(){
```
nobodyiam: I think this is not necessary, as we could use `com.ctrip.framework.apollo.model.ConfigChangeEvent#getChange`

apollo-client/pom.xml
```diff
@@ -11,7 +11,7 @@
 	<artifactId>apollo-client</artifactId>
 	<name>Apollo Client</name>
 	<properties>
-		<java.version>1.7</java.version>
+		<java.version>1.8</java.version>
```
nobodyiam: I don't think it's necessary to change the target version from 1.7 to 1.8 simply for this dynamic log level feature

apollo-client/src/main/java/com/ctrip/framework/apollo/listener/LogLevelListener.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.stereotype.Service;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Created by lawrence on 19/12/16.
+ */
+@Service
```
nobodyiam: This feature relies on Spring Boot, so I think we should create a new module like `apollo-spring-boot-starter` to put Spring Boot related features.

And for each feature, we could provide a flag so that users can decide to enable it or not. We could achieve this via Spring Boot AutoConfiguration features.

lawrencewu: I agree. And this feature can resolve the dynamic log level change with the "LoggingSystem" help which class  in Spring Boot, so that we do not need to redefine an interface to archive this goal under spring project.

apollo-client/pom.xml
lawrencewu: ok

apollo-client/src/main/java/com/ctrip/framework/apollo/model/ConfigChangeEvent.java
lawrencewu: sure. ok

PR #2861, opened 2019-12-23
Use OrderedProperties instead of Properties (#2)

### LINKED ISSUE #2451: 《使用apollo配置zuul动态路由不能保证数据有序性的问题》

songdragon:

Keep order of configuration in yml/yaml at client side.

Fixes #2451, #2273

## Brief changelog

* replace Properties implemantion with OrderedProperties
* override put and get from Hashtable
* add two order test case
* add comments for overrided methods


songdragon:
@nobodyiam when fixing test failures, I found that if someone use Properties API,  there are more work to do, i.e. to override all Properties API in OrderedProerties. This PR only implemented APIs (e.g.: put/get) to pass tests. 

It is better to remind in Wiki or somewhere if this PR is accepted.

nobodyiam:
@songdragon 

Thanks for your quick response, this PR now contains less changes and is more clear. 

However, your reminder `This PR only implemented APIs (e.g.: put/get) to pass tests.` might be a problem which may introduce compatibility issues in the future.

songdragon:
@nobodyiam 
Yes, it's definitely true. I am working on checking unimplemented methods in OrderedProperties. If there is anything new, I wil update this PR.

songdragon:
Listed all public methods in Properties and Hashtable, and unimplemented methods in OrderedProperties are labed **todo**.
[x] means implemented

|Methods|OrderedProperties| Properties|Hashtable|
| ------------- | ------------- | ------------- | ------------- |
| clear():void  | todo(1st)  |inherited|[x]|
| clone():void  | todo(1st) | inherited| [x]|
| computeIfAbsent(K, Function(? super K, ? extends V>:V  | todo  |inherited|[x]| 
| computeIfPresent(K, BiFunction(? super K,? super V, ? extends V>:V  | todo  |inherited|[x]| 
| compute(K, BiFunction(? super K, ? super V, ? extends V>:V | todo  |inherited|[x]| 
| contains(Object):boolean  | [x]  |[x]||
| containsValue(Object):boolean  | [x]  |[x]||
| elements():Enumeration<V>  | todo(1st)  |[x]||
| entrySet():Set<Map.Entry<K,V>>  | [x]  |inherited|[x]|
| equals(Object):boolean  | [x]  |inherited|[x]|
| forEach(BiConsumer<? super K, ? super V>):void  | todo  |inherited|[x]|
| get(Object): V  | [x]  |inherited|[x]|
| getOrDefault(Object,V):V  | todo(1st)  |inhertied|[x]|
| getProperty(String):String  | [x]  |[x]|
| getProperty(String,String):String  | [x]  |[x]|
| hashCode():int  | [x]  |inherited|[x]|
| isEmpty():boolean  | [x]  |[x]|[x]|
| keys():Enumeration<K>  |todo(1st) |inherited | [x]|
| keySet(): Set<K>  | [x]  |inherited|[x]|
| list(PrintStream):void  | [x]  |[x]|
| list(PrintWriter):void  | [x]  |[x]|
| load(InputStream):void | [x]  |[x] ||
| load(Reader):void  | [x]  |[x] ||
| loadFromXML(InputStream):void  | [x]  |[x]|
| merge(K, BiFunction(? super K, ? super V, ? extends V>:V  | todo  |inherited|[x]| 
| put(K,V)  | [x]  |inherited|[x]|
| putAll(Map<? extends K,?extends V):void  | [x]  | inherited |[x]|
| putIfAbsent(K,V):V  | todo  |inherited|[x]| 
| propertyNames():Enumeration<?>  | [x]  |[x]|
| remove(Object):V  | todo(1st)  |inherited|[x]|
| remove(Object,Object):boolean  | todo  |inherited|[x]| 
| replace(K,V,V):boolean  | todo(1st)  |inherited|[x]| 
| replace(K,V):V | todo(1st)  |inherited|[x]| 
| replaceAll(BiFunction<? super K, super V,? extends V>):void  | todo   |inherited|[x]|
| size():int  | [x]  |[x]|[x]|
| setProperty(String,String):Object  | [x]  |[x]||
| save(OutputStream,String):void  | Not required |[x]| |
| store(Writer,String):void  | [x]  |[x]||
| store(OutputStream,String):void  | [x]  |[x]||
| storeToXML(OutputStream,String):void  | [x]  |[x]||
| storeToXML(OutputStream,String,String):void  | [x]  |[x]||
| stringPropertyNames():Set<String>  | [x]  |[x]||
| toString():String  | [x]  |inherited|[x]|
| values(): Collection<V>  | todo(1st)  |inherited|[x]|

songdragon:
@nobodyiam 
Code has been updated to implement Properties and Hashtable methods in JDK1.7 as above. However `Hashtable in JDK1.8 has new APIs/Stream APIs`, which are not compitable in target JDK1.7. These methods are implemented but commented in source code. They should be commented if client module is compiled to 1.8 and above. 

Do you have any idea to solve this compatibility issues?

nobodyiam:
@songdragon 

I'm wondering do we really need a fully functional `OrderedProperties` to fix #2451, #2273?

I haven't tested it yet, but I guess if we can make `com.ctrip.framework.apollo.Config#getPropertyNames` return the correctly ordered property names, then these 2 issues will be fixed?

If this is the case, then we may only need a simplified `OrderedProperties` like 

```java
public class OrderedProperties extends Properties {

  private final LinkedHashSet<String> propertyNames;

  public OrderedProperties() {
    propertyNames = new LinkedHashSet<>();
  }

  @Override
  public synchronized Object put(Object key, Object value) {
    addPropertyName(key);
    return super.put(key, value);
  }

  @Override
  public Set<String> stringPropertyNames() {
    return propertyNames;
  }

  @Override
  public Enumeration<?> propertyNames() {
    return Collections.enumeration(propertyNames);
  }

  @Override
  public synchronized Enumeration<Object> keys() {
    return new Enumeration<Object>() {
      private final Iterator<String> i = propertyNames.iterator();

      public boolean hasMoreElements() {
        return i.hasNext();
      }

      public Object nextElement() {
        return i.next();
      }
    };
  }

  @Override
  public Set<Object> keySet() {
    return new LinkedHashSet<Object>(propertyNames);
  }

  @Override
  public Set<Entry<Object, Object>> entrySet() {
    Set<Entry<Object, Object>> original = super.entrySet();
    LinkedHashMap<Object, Entry<Object, Object>> entryMap = new LinkedHashMap<>();
    for (String propertyName : propertyNames) {
      entryMap.put(propertyName, null);
    }

    for (Entry<Object, Object> entry : original) {
      entryMap.put(entry.getKey(), entry);
    }

    return new LinkedHashSet<>(entryMap.values());
  }

  private void addPropertyName(Object key) {
    if (key instanceof String) {
      propertyNames.add((String) key);
    }
  }
}
```

How do you think?

songdragon:
@nobodyiam 
Agree with you. It's really a smart way and it works well.  

However, I am wondring there may be some inconsistency behavier. E.g.: clear() only clean storage but not clear ordered key names, but this is not required to fix #2451, #2273.

To solve the order issue, there are two key points:    
1. keep order in storage data structue, i.e. the `OrderedProperties`
2. keep order in temproral useage, e.g.  stringPropertyNames()  in `DefaultConfig`

nobodyiam:
@songdragon 

By using this way, I think it's more simple and clear. Regarding the `inconsistency behavier` you mentioned, I think we can support it, i.e. to implement the clear, remove, putIfAbsent, etc.

BTW, have you tested the scenarios of #2451, #2273? I think we need to add some integration cases to make sure the problem of #2451, #2273 is solved and make sure it won't break due to some unexpected changes in the future.

songdragon:
@nobodyiam 
Added 3 test cases in ConfigIntegrationTest:
1. No local file with remote config
2. Both local file and remote config
3. Local file with remote config error

For the `inconsistency behavier` issue, putAll,clear and remove methods are implemented under 1.7 target. I think 1.8/steram APIs are only required if we target client module to only run on JDK1.8+.

apollo-client/src/main/java/com/ctrip/framework/apollo/util/OrderedProperties.java
```diff
@@ -79,4 +73,25 @@ private void addPropertyName(Object key) {
       propertyNames.add((String) key);
     }
   }
+
+  @Override
+  public synchronized void putAll(Map<?, ?> t) {
```
nobodyiam: I think `putAll` is unnecessary?

songdragon: Sorry for running tests on JDK10, and add this method to fix tests failure. Just notice that Properties are different in JDK1.8/1.7 and JDK 10. The `putAll` method is inherited from `Properites` class in JDK 10 while it is from `Hashtable` in JDK1.8/1.7. 
Remove it in lastest commit.

songdragon: This method is required in JDK10, due to code in `RemoteConfigRepository` L163 invoking putAll.

``` JAVA
  private Properties transformApolloConfigToProperties(ApolloConfig apolloConfig) {
    Properties result =  PropertiesFactory.getPropertiesObject();     
    result.putAll(apolloConfig.getConfigurations());
    return result;
  }
```

nobodyiam: You are right, the implementation of `Properties` in JDK 10 is totally different from that in JDK 7/8...So `putAll` is required.

And for this feature, it is important we test it against JDK 7, 8, 11.

songdragon: I think apollo is not ready for JDK 11, because UT will fail on JDK 11. 
So should we add putAll now, or add it until all are prepared for JDK11?

nobodyiam: Actually apollo is ok with JDK 11, which we have tested functionally.
(There is one issue brought by #2828, which is fixed in #2888)

However, the UT does fail with JDK 11 and I will try to fix it later.

So in a word, we need to test the function againt JDK 7, 8 and 11.

apollo-client/src/main/java/com/ctrip/framework/apollo/internals/AbstractConfig.java
```diff
@@ -493,11 +495,11 @@ private boolean isConfigChangeListenerInterested(ConfigChangeListener configChan
   List<ConfigChange> calcPropertyChanges(String namespace, Properties previous,
                                          Properties current) {
     if (previous == null) {
-      previous = new Properties();
+      previous =  PropertiesFactory.getPropertiesObject();
```
nobodyiam: I think we need to use `ApolloInjector` to get the PropertiesFactory instance so that it is testable.

nobodyiam: btw, there is one more extra space before `PropertiesFactory`

apollo-client/src/main/java/com/ctrip/framework/apollo/util/OrderedProperties.java
```diff
@@ -0,0 +1,112 @@
...
+      entryMap.put(entry.getKey(), entry);
+    }
+
+    return new LinkedHashSet<>(entryMap.values());
+  }
+
+  @Override
+  public synchronized void putAll(Map<?, ?> t) {
+    for(Entry entry:t.entrySet()){
+      put(entry.getKey(),entry.getValue());
```
nobodyiam: how about

```java
  @Override
  public synchronized void putAll(Map<?, ?> t) {
    super.putAll(t);
    for (Object key : t.keySet()) {
      addPropertyName(key);
    }
  }
```

apollo-client/src/main/java/com/ctrip/framework/apollo/util/factory/PropertiesFactory.java
```diff
@@ -0,0 +1,11 @@
+package com.ctrip.framework.apollo.util.factory;
+
+import com.ctrip.framework.apollo.util.OrderedProperties;
+import java.util.Properties;
+
+public class PropertiesFactory {
+
+  public static Properties getPropertiesObject(){
+    return new OrderedProperties();
```
nobodyiam: How about we adding a flag here to enable this ordered properties feature only if user intends to?

songdragon: Good idea! Implemented in latest commits.

apollo-client/src/main/java/com/ctrip/framework/apollo/util/OrderedProperties.java
songdragon: More clear, `OrderedProerites` only maintains keys order and `Properties` is responsible for storage keys/values.

apollo-client/src/main/java/com/ctrip/framework/apollo/internals/AbstractConfig.java
songdragon: Already changed in latest commits. Please tell me  if  I use `ApolloInjector` in a wrong way.

apollo-client/src/main/java/com/ctrip/framework/apollo/internals/AbstractConfig.java
```diff
@@ -68,6 +71,7 @@ public AbstractConfig() {
     m_configVersion = new AtomicLong();
     m_arrayCache = Maps.newConcurrentMap();
     allCaches = Lists.newArrayList();
+    propertiesFactory=ApolloInjector.getInstance(PropertiesFactory.class);
```
nobodyiam: please format this line, i.e `propertiesFactory = ApolloInjector.getInstance(PropertiesFactory.class);`

apollo-client/src/main/java/com/ctrip/framework/apollo/util/yaml/YamlParser.java
```diff
@@ -25,12 +27,14 @@
 public class YamlParser {
   private static final Logger logger = LoggerFactory.getLogger(YamlParser.class);
 
+  private PropertiesFactory propertiesFactory= ApolloInjector.getInstance(PropertiesFactory.class);
```
nobodyiam: please format this line too

apollo-client/src/main/java/com/ctrip/framework/apollo/internals/AbstractConfig.java
nobodyiam: It looks OK.

apollo-client/src/main/java/com/ctrip/framework/apollo/util/factory/DefaultPropertiesFactory.java
```diff
@@ -0,0 +1,42 @@
...
+  private static final Logger logger= LoggerFactory.getLogger(DefaultPropertiesFactory.class);
+
+  protected boolean isOrderEnabled=false;
+
+  public DefaultPropertiesFactory(){
+    initOrderEnabled();
+  }
+
+  private void initOrderEnabled() {
+    String customizedOrderEnable = System.getProperty(APOLLO_PROPERTY_ORDER_ENABLE);
```
nobodyiam: I think we could support setting this flag in more ways. 

e.g. 

1. app.properties
```java
Foundation.app().getProperty(APOLLO_PROPERTY_ORDER_ENABLE, null);
```

2. Spring environment
Just add it to `com.ctrip.framework.apollo.spring.boot.ApolloApplicationContextInitializer#APOLLO_SYSTEM_PROPERTIES`

https://github.com/ctripcorp/apollo/blob/42edb9da963df986538edcda4ff58d474fb3abed/apollo-client/src/main/java/com/ctrip/framework/apollo/spring/boot/ApolloApplicationContextInitializer.java#L63

apollo-client/src/main/java/com/ctrip/framework/apollo/util/yaml/YamlParser.java
nobodyiam: there are some more lines with the similar issue, please also check and formar them

songdragon:
@nobodyiam 
Follwing things has been updated, please review them again.    
1. Code has been reformated using intellij-java-google-style.xml. 
2. `APOLLO_PROPERTY_ORDER_ENABLE` flag is assigned in app.properties or spring environment.

PR #2864, opened 2019-12-24
Refactor/remove redundant else

waters00:
## What's the purpose of this PR

remove redundant else.
If an if block contains a return/throw statement, the else block becomes unnecessary. Its contents can be placed outside of the block.

## Which issue(s) this PR fixes:
Fixes #

## Brief changelog

refactor code.

Follow this checklist to help us incorporate your contribution quickly and easily:

- [x] Read the [Contributing Guide](https://github.com/ctripcorp/apollo/blob/master/CONTRIBUTING.md) before making this pull request.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Write necessary unit tests to verify the code.
- [x] Run `mvn clean test` to make sure this pull request doesn't break anything.

PR #2867, opened 2019-12-24
Dynamic environment for apollo-portal

Anilople:
What's the purpose of this PR

Add a **feature**, make portal support self-define environment without change 1 line code.
Only config will be changed.

## Brief changelog

Dynamic environment setting when portal service start.
A new way to forbidden changing hard code to add new environment.
Just change 2 configuration properties.
Suppose that you want to add a new environment **beta**, and you have been add **dev**, just do like follow step.

1. change `apollo.portal.envs`.Just change its value to `dev,beta` in db or add VM option `-Dapollo.portal.envs=dev,beta`
2. add property `beta.meta=your-meta-address` or add VM option `-Dbeta_meta=your-meta-address`

There are 4 fake classes added under package `com.ctrip.framework.apollo.portal.environment`,
the purpose is using them to replace the classes in `apollo-core`.

But only 1 fake class exists testing, the rest 3 classes lack the unit testing.

Could I cancel the function `3rd party MetaServerProvider could be injected by typical Java Service Loader pattern.` in `com.ctrip.framework.apollo.portal.environment.PortalMetaDomainConsts`, and make its loading to be not lazy ?


apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalLegacyMetaServerProvider.java
```diff
@@ -0,0 +1,97 @@
...
+            String value = entry.getValue();
+            // put pair (Env, meta server address)
+            domains.put(env, value);
+        }
+
+        // log all
+        logger.info("All environment's meta server address: {}", domains);
+    }
+
+    private String getMetaServerAddress(Properties prop, String sourceName, String propName) {
```
nobodyiam: It seems this method is not used.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaDomainConsts.java
```diff
@@ -0,0 +1,198 @@
...
+    public static String getMetaServerAddress(Env env) {
+        if (!metaServerAddressCache.containsKey(env)) {
+            initMetaServerAddress(env);
+        }
+
+        return metaServerAddressCache.get(env);
+    }
+
+    private static void initMetaServerAddress(Env env) {
+        if (metaServerProviders == null) {
```
nobodyiam: I think we don't need this  meta server provider spi for portal, so we could simplify this logic.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/service/ClusterService.java
```diff
@@ -26,7 +26,7 @@ public ClusterService(final UserInfoHolder userInfoHolder, final AdminServiceAPI
     return clusterAPI.findClustersByApp(appId, env);
   }
 
-  public ClusterDTO createCluster(Env env, ClusterDTO cluster) {
+  public synchronized ClusterDTO createCluster(Env env, ClusterDTO cluster) {
```
nobodyiam: I don't think adding `synchronized` here will solve the concurrency problem because there may be more than 1 apollo-portal instances running. That should be solved in apollo-adminservice, probably with a distributed lock.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalLegacyMetaServerProvider.java
Anilople: Yes. This method is the old way to get meta server address. After using new way to get meta server address, this method is not using any more.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaDomainConsts.java
Anilople: Initialize `metaServerProviders` without lazy? i.e initial it when class loaded? (use `final` to the field)

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/service/ClusterService.java
Anilople: Emmm. It seems like that. I will change the code back.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaDomainConsts.java
Anilople: > don't need this meta server provider spi for portal

It mean that there are no need to use the `com.ctrip.framework.apollo.portal.environment.PortalMetaServerProvider`, simply cancel this interface, and use `com.ctrip.framework.apollo.portal.environment.PortalLegacyMetaServerProvider`  directly?

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalLegacyMetaServerProvider.java
nobodyiam: Then we may just remove it.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaDomainConsts.java
nobodyiam: I think we could remove the PortalMetaServerProvider spi for now

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalLegacyMetaServerProvider.java
```diff
@@ -0,0 +1,97 @@
...
+import com.google.common.base.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+
+/**
+ * a fake class to replace com.ctrip.framework.apollo.core.internals.LegacyMetaServerProvider in apollo-portal
```
nobodyiam: This is not a fake class, it's just an implementation for apollo-portal only, so I think we could change the description a little bit

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaDomainConsts.java
```diff
@@ -0,0 +1,198 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * a fake class only use in apollo-portal
```
nobodyiam: This is not a fake class, it's just an implementation for apollo-portal only, so I think we could change the description a little bit

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalLegacyMetaServerProvider.java
Anilople: Got it.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaDomainConsts.java
Anilople: Ok. I will remove the interface

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalLegacyMetaServerProvider.java
Anilople: Get it.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaDomainConsts.java
Anilople: Get it.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/controller/ConsumerController.java
```diff
@@ -86,32 +74,33 @@ public ConsumerToken getConsumerTokenByAppId(@RequestParam String appId) {
...
-      throw new BadRequestException("Params(NamespaceName) can not be empty.");
-    }
-    if (null != envs){
-      String[] envArray = envs.split(",");
-      List<String> envList = Lists.newArrayList();
-      // validate env parameter
-      for (String env : envArray) {
-        if (Strings.isNullOrEmpty(env)) {
-          continue;
+    } else {
```
nobodyiam: I think this `else {` statement is unnecessary

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalLegacyMetaServerProvider.java
```diff
@@ -0,0 +1,79 @@
...
+ * Only use in apollo-portal
+ * load all meta server address from
+ *  - System Property           [key ends with "_meta"]
+ *  - OS environment variable   [key ends with "_meta"]
+ *  - user's configuration file [key ends with ".meta"]
+ * when apollo-portal start up.
+ * @see com.ctrip.framework.apollo.core.internals.LegacyMetaServerProvider
+ * @author wxq
+ */
+public class PortalLegacyMetaServerProvider {
```
nobodyiam: How about just `PortalMetaServerProvider`?

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalLegacyMetaServerProvider.java
```diff
@@ -0,0 +1,79 @@
...
+            String value = entry.getValue();
+            // put pair (Env, meta server address)
+            domains.put(env, value);
+        }
+
+        // log all
+        logger.info("All environment's meta server address: {}", domains);
+    }
+
+    public String getMetaServerAddress(Env targetEnv) {
```
nobodyiam: Please add some unit tests to cover the meta server address logic

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaDomainConsts.java
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Only use in apollo-portal
+ * Provider an available meta server url.
+ * If there is no available meta server url for the given environment,
+ * the default meta server url will be used(http://apollo.meta).
+ * @see com.ctrip.framework.apollo.core.MetaDomainConsts
+ * @author wxq
+ */
+public class PortalMetaDomainConsts {
```
nobodyiam: Please add unit test for PortalMetaDomainConsts

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalLegacyMetaServerProvider.java
Anilople: Ok

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalLegacyMetaServerProvider.java
Anilople: Ok

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaDomainConsts.java
Anilople: In `apollo-core`, there is a test for `MetaDomain`, i.e `com.ctrip.framework.apollo.core.MetaDomainTest`. It seems that implements `com.ctrip.framework.apollo.BaseIntegrationTest`, and start a internal  server to test.

If i want to write the unit test of `com.ctrip.framework.apollo.portal.environment.PortalMetaDomainConsts`, should i use the same way?

A little Challenge to do this. (hard)

nobodyiam: I think we need to do it and it looks like you have done that part

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/controller/ConsumerController.java
Anilople: From the git history. The code is merged from other branch.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaDomainConsts.java
Anilople: The unit test is done.

Anilople: Most code is copied from `apollo-core`

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/controller/ConsumerController.java
nobodyiam: the `else {` statement was removed in pr #2864 , so I guess it was brought back to this branch when you resolved the merging conflicts.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaServerProvider.java
```diff
@@ -0,0 +1,131 @@
...
+     * load all environment's meta address dynamically when this class loaded by JVM
+     */
+    private static Map<Env, String> initializeDomains() {
+        // find key-value from System Property which key ends with "_meta"
+        Map<String, String> metaServerAddressesFromSystemProperty = KeyValueUtils.filterWithKeyEndswith(System.getProperties(), "_meta");
+        // remove key's suffix "_meta"
+        metaServerAddressesFromSystemProperty = KeyValueUtils.removeKeySuffix(metaServerAddressesFromSystemProperty, "_meta".length());
+
+        // find key-value from OS environment variable which key ends with "_meta"
+        Map<String, String> metaServerAddressesFromOSEnvironment = KeyValueUtils.filterWithKeyEndswith(System.getenv(), "_meta");
```
nobodyiam: OS Environment variables are normally in UPPER CASE, so it might be better to check with `_META`

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaServerProvider.java
```diff
@@ -0,0 +1,131 @@
...
+        // remove key's suffix "_meta"
+        metaServerAddressesFromOSEnvironment = KeyValueUtils.removeKeySuffix(metaServerAddressesFromOSEnvironment, "_meta".length());
+
+        // find key-value from properties file which key ends with ".meta"
+        Properties properties = new Properties();
```
nobodyiam: I think it might be better to also support setting meta server address from PortalDB.ServerConfig, which is the same place to store `apollo.portal.envs` and is therefore more intuitive, e.g. user can set env and env meta server address in `System Configuration` page. How do you think?

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/controller/ConsumerController.java
Anilople: Change back already.

nobodyiam: Is the whole file rolled back? I think we still need to change the import statement `import com.ctrip.framework.apollo.portal.environment.Env;`?

Anilople: I've changed it by accident. A new commit has pushed.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaServerProvider.java
Anilople: I've change the rule of match suffix to case insensitive.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/environment/PortalMetaServerProvider.java
Anilople: You mean the meta server addresses are save in `apollo.portal.envs` which type is a `map`?
the config like follow:
```yml
apollo:
  portal:
    envs:
      dev: http://ip:port
      other: http://ip:port
      self-define: http://ip:port
```
And in `PortalDB.ServerConfig`'s `apollo.portal.envs`, there is a json object
```
{
      "dev"" : "http://ip:port"
      "other" : "http://ip:port"
      "self-define" : "http://ip:port"
}
```
So then portal conbine those config to get the meta server addresses?

Anilople: I think this feature (add environment in portal when running) will consume more code to implement.
How about when this feature (add environment in portal when start up) finished, next plan to do it?

nobodyiam: I think we can keep it as `add environment in portal when start up`, but we support configuring it via `System Configuration` page, e.g. we add a new configuration item like `apollo.portal.meta.servers` to store the mapping between env and meta server address.

We will still keep the `apollo.portal.envs` as it indicates the active environments so that there may be 3 environment mappings in `apollo.portal.meta.servers` but only 2 active envs in `apollo.portal.envs`.

If this involves much more effort, then it's okay to leave it when we support `add environment in portal when running`.

Anilople: Get it. 
It involves the Database operations. Add new key `apollo.portal.meta.servers`, and refresh it in runtime.
I feel that the database operation is a bit difficult.
I don't know how to add this feature now.

PR #3144, opened 2020-06-24
update NotificationControllerV2.java

jiangcaijun:

constructor （for set and map ）: choose more appropriate initialCapacity

### DIFF
--- a/apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/controller/NotificationControllerV2.java
+++ b/apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/controller/NotificationControllerV2.java
@@ -38,11 +38,7 @@ import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.context.request.async.DeferredResult;
 
 import java.lang.reflect.Type;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/controller/NotificationControllerV2.java
```diff
@@ -111,10 +107,14 @@ public NotificationControllerV2(
     }
 
     DeferredResultWrapper deferredResultWrapper = new DeferredResultWrapper(bizConfig.longPollingTimeoutInMilli());
-    Set<String> namespaces = Sets.newHashSet();
-    Map<String, Long> clientSideNotifications = Maps.newHashMap();
+   
     Map<String, ApolloConfigNotification> filteredNotifications = filterNotifications(appId, notifications);
 
+    //appoint initialCapacity
```
nobodyiam, line 113: how about using the existing methods?

```java
    Set<String> namespaces = Sets.newHashSetWithExpectedSize(filteredNotifications.size());
    Map<String, Long> clientSideNotifications = Maps.newHashMapWithExpectedSize(filteredNotifications.size());
```

jiangcaijun: it is great。

I didn't think of this idea， because  I don't use `guava` very often

jiangcaijun: so, would you commit this code by yourself ?

At the same time, i find some other optimization 。because this `NotificationControllerV2.java` is so important that it cannot be too concerned

the PR is : #3145

jiangcaijun:
combine #3144 to #3145 , as one pr

PR #3145, opened 2020-06-25
Update NotificationControllerV2.java

jiangcaijun:
##  What's the purpose of this PR

Adjust the order of check。

if `filteredNotifications` is not empty, then  `namespaces` will not empty。 so check  `filteredNotifications` before  `namespaces ` Initialized ， will avoid create useless object


nobodyiam:
I think it's ok to combine #3144 and #3145 as one pr

jiangcaijun:
> I think it's ok to combine #3144 and #3145 as one pr

done。please check

PR #3380, opened 2020-11-15
对代码进行部分优化以及添加代码注释

smilesnake:
代码进行部分优化以及添加代码注释


nobodyiam (changes requested):
Thanks for your contribution, here are some comments:
1. Please try ot keep the comment as English, as it is more user-friendly in the long term
2. Please do not use lombok in a widely used open source project, as it will force other users to install the lombok plugin or use an IDE
3. Please submit different pull request for different purposes, e.g. one pull request to adding comments, one pull request to adjust the format, and one pull request to refactor some logics

PR #3573, opened 2021-02-26
apollo client shared object context

vdisk-group:
1. add a shared object context as a common way to customize apollo client
2. customize http client by the shared object context

### DIFF
--- a/apollo-client/src/main/java/com/ctrip/framework/apollo/internals/DefaultInjector.java
+++ b/apollo-client/src/main/java/com/ctrip/framework/apollo/internals/DefaultInjector.java
@@ -1,6 +1,9 @@
 package com.ctrip.framework.apollo.internals;
 
 import com.ctrip.framework.apollo.exceptions.ApolloConfigException;
+import com.ctrip.framework.apollo.shared.ApolloSharedContext;
+import com.ctrip.framework.apollo.shared.ApolloSharedContextImpl;
+import com.ctrip.framework.apollo.spi.ApolloSharedContextCustomizer;
 import com.ctrip.framework.apollo.spi.ConfigFactory;
 import com.ctrip.framework.apollo.spi.ConfigFactoryManager;
 import com.ctrip.framework.apollo.spi.ConfigRegistry;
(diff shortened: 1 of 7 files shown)


nobodyiam:
This is a good try. However, I've been thinking whether we could provide a more general solution so that users could customize other instances as well.

Here is one of my thought, please take a look and see if this works.

1. Define an interface `ApolloInjectorCustomizer` like below:

```java
package com.ctrip.framework.apollo.spi;

import com.ctrip.framework.apollo.core.spi.Ordered;
import com.ctrip.framework.apollo.internals.Injector;

public interface ApolloInjectorCustomizer extends Injector, Ordered {

}
```

2. Modify the implementation of `DefaultInjector` as below:

```java
public class DefaultInjector implements Injector {
  private com.google.inject.Injector m_injector;
  private List<ApolloInjectorCustomizer> customizers;

  public DefaultInjector() {
    try {
      m_injector = Guice.createInjector(new ApolloModule());
      customizers = ServiceBootstrap.loadAllOrdered(ApolloInjectorCustomizer.class);
    } catch (Throwable ex) {
      ApolloConfigException exception = new ApolloConfigException("Unable to initialize Guice Injector!", ex);
      Tracer.logError(exception);
      throw exception;
    }
  }

  @Override
  public <T> T getInstance(Class<T> clazz) {
    try {
      for (ApolloInjectorCustomizer customizer : customizers) {
        T instance = customizer.getInstance(clazz);
        if (instance != null) {
          return instance;
        }
      }
      return m_injector.getInstance(clazz);
    } catch (Throwable ex) {
      Tracer.logError(ex);
      throw new ApolloConfigException(
          String.format("Unable to load instance for %s!", clazz.getName()), ex);
    }
  }

  @Override
  public <T> T getInstance(Class<T> clazz, String name) {
    for (ApolloInjectorCustomizer customizer : customizers) {
      T instance = customizer.getInstance(clazz, name);
      if (instance != null) {
        return instance;
      }
    }
    //Guice does not support get instance by type and name
    return null;
  }

  private static class ApolloModule extends AbstractModule {
    @Override
    protected void configure() {
      bind(ConfigManager.class).to(DefaultConfigManager.class).in(Singleton.class);
      bind(ConfigFactoryManager.class).to(DefaultConfigFactoryManager.class).in(Singleton.class);
      bind(ConfigRegistry.class).to(DefaultConfigRegistry.class).in(Singleton.class);
      bind(ConfigFactory.class).to(DefaultConfigFactory.class).in(Singleton.class);
      bind(ConfigUtil.class).in(Singleton.class);
      bind(HttpUtil.class).to(DefaultHttpClient.class).in(Singleton.class);
      bind(ConfigServiceLocator.class).in(Singleton.class);
      bind(RemoteConfigLongPollService.class).in(Singleton.class);
      bind(YamlParser.class).in(Singleton.class);
      bind(PropertiesFactory.class).to(DefaultPropertiesFactory.class).in(Singleton.class);
    }
  }
}
```

> we need to change the `com.ctrip.framework.foundation.internals.ServiceBootstrap#loadAllOrdered` to allow empty list

3. Users can now offer their customized solution as below

```java
public class SomeCustomizer implements ApolloInjectorCustomizer {

  @Override
  public <T> T getInstance(Class<T> clazz) {
    // do some custom logic here
    return null;
  }

  @Override
  public <T> T getInstance(Class<T> clazz, String name) {
    return null;
  }

  @Override
  public int getOrder() {
    return 0;
  }
}
```

nobodyiam:
@vdisk-group what's your idea on this? I think it's quite useful to have this feature.

vdisk-group:
> @vdisk-group what's your idea on this? I think it's quite useful to have this feature.

It looks great, I hope it will be available on the next version

PR #3593, opened 2021-03-10
ServiceBootstrap unit test fix

vdisk-group:
on jdk11 the method ```java.util.ServiceLoader#load```  called the check method ```java.util.ServiceLoader#checkCaller```
the ```private interface``` cause the error ```service type not accessible to xxx```
```
    private static void checkCaller(Class<?> caller, Class<?> svc) {
        ...
        // Check access to the service type
        Module callerModule = caller.getModule();
        int mods = svc.getModifiers();
        if (!Reflection.verifyMemberAccess(caller, svc, null, mods)) {
            fail(svc, "service type not accessible to " + callerModule);
        }
        ...
    }

```

### DIFF
--- a/apollo-core/src/test/java/com/ctrip/framework/foundation/internals/ServiceBootstrapTest.java
+++ b/apollo-core/src/test/java/com/ctrip/framework/foundation/internals/ServiceBootstrapTest.java
@@ -48,25 +48,25 @@ public class ServiceBootstrapTest {
     ServiceBootstrap.loadPrimary(Interface7.class);
   }
 
-  public interface Interface1 {
+  interface Interface1 {
   }
 
   public static class Interface1Impl implements Interface1 {
   }
 
-  public interface Interface2 {
+  interface Interface2 {
   }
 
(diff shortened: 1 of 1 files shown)


apollo-core/src/test/java/com/ctrip/framework/foundation/internals/ServiceBootstrapTest.java
```diff
@@ -48,25 +48,25 @@ public void loadPrimaryWithServiceFileButNoServiceImpl() {
     ServiceBootstrap.loadPrimary(Interface7.class);
   }
 
-  private interface Interface1 {
+  public interface Interface1 {
```
nobodyiam, line 51: How about this?
```suggestion
  interface Interface1 {
```

vdisk-group: That's right, package-private is enough for ServiceLoader

nobodyiam (approved):
LGTM

PR #3594, opened 2021-03-10
replace http client implementation with interface

vdisk-group:
make it easier to customize http client


apollo-client/src/main/java/com/ctrip/framework/apollo/util/http/DefaultHttpClient.java
```diff
@@ -19,14 +19,14 @@
 /**
  * @author Jason Song(song_s@ctrip.com)
  */
-public class HttpUtil {
```
nobodyiam, line 22: I think we may still keep the class `HttpUtil` and delegate its implementation to the `HttpClient` because some application might already reference it. I did a [quick search](https://github.com/qw7575/spring-boot-examples/blob/9dd0b838262ea86c018684cb6218c1540039e64a/spring-boot-jsoup/src/main/java/com/qw/boot/examples/jsoup/service/LogicServiceImpl.java) in github, and found the possibility is true.

To avoid future misuse, we could mark it as `@Deprecated`.

nobodyiam (changes requested):

apollo-client/src/main/java/com/ctrip/framework/apollo/util/http/HttpUtil.java
```diff
@@ -18,8 +18,10 @@
 
 /**
  * @author Jason Song(song_s@ctrip.com)
+ * @deprecated in favor of the interface {@link HttpClient} and it's default implementation {@link DefaultHttpClient}
  */
-public class HttpUtil {
+@Deprecated
+public class HttpUtil implements HttpClient {
```
nobodyiam, line 24: How about we delegate the implementation to DefaultHttpClient?

```java
  public HttpUtil() {
    m_httpClient = new DefaultHttpClient();
  }

  @Override
  public <T> HttpResponse<T> doGet(HttpRequest httpRequest, final Class<T> responseType) {
    return m_httpClient.doGet(httpRequest, responseType);
  }
```

PR #3628, opened 2021-04-04
fix oidc logout

### LINKED ISSUE #3625: Username is not corrected with ocdi user's subject
we are integrating Apollo with Keycloak (openid provider), and got below issues:

1. it seems to be not corrected that Apollo use ocdi user's subject as user id , should be some other properties such as email/username... because the subject of ocdi user's subject may be ID in database. it's better that can be configurated by user.
2. if the apollo work on Nginx or with some context path such as apollo, when logout (integrate with openid), it redirect to /logout, should be /apollo/logout.

vdisk-group:

fix oidc logout with context-path

Fixes #3625


nobodyiam (approved):
LGTM

PR #3667, opened 2021-05-08
test(apollo-client): use assertEquals instead of assertThat

Anilople:

Fix test fail when use java 9 or above.

NONE

## Brief changelog

Use equals to compare properties directly.

## details

Consider code like follow
```java
import java.util.Properties;

public class Main {

  public static void main(String[] args) {
    Properties a = new Properties();
    a.put("x", "0");

    Properties b = new Properties();
    b.put("x", "0");

    System.out.println(a.equals(b));
    System.out.println(a.entrySet().equals(b.entrySet()));
  }
}
```

In Jdk 8, the output will be
```
true
true
```

But in Jdk 9 or above, the output will change to
```
true
false
```

One can run it on website https://www.jdoodle.com/online-java-compiler/

So it need to compare properties directly instread of using entrySet.


nobodyiam (approved):
LGTM

PR #3669, opened 2021-05-09
fix unit test

vdisk-group:
newSingleThreadExecutor has a bug, and it only fixed on jdk11. while running on jdk8, it may break the unit test by accident (depend on gc)
https://bugs.openjdk.java.net/browse/JDK-8145304


nobodyiam (approved):
LGTM

PR #3677, opened 2021-05-11
解决日志系统未初始化完成时，apollo 的加载日志没法输出问题

klboke:

> 解决问题：

解决 spring boot 方式集成时，配置  `apollo.bootstrap.eagerLoad.enabled` 情况下，apollo 的加载日志没法输出问题。

> 设计原理：

封装一个基于 guava-cache 的日志输出内容存储容器 `DeferredLogUtil`，将 apollo 初始化时的日志信息，添加到延迟输出的日志容器。在日志系统加载完成也就是 EnvironmentPostProcessor 事件立刻回放，最终实现的效果和不开启 `apollo.bootstrap.eagerLoad.enabled` 是一样的，设计过程参考了 [spring-boot#DeferredLog](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/DeferredLog.java)


apollo-client/src/main/java/com/ctrip/framework/apollo/internals/DefaultConfig.java
```diff
@@ -58,8 +59,9 @@ private void initialize() {
       updateConfig(m_configRepository.getConfig(), m_configRepository.getSourceType());
     } catch (Throwable ex) {
       Tracer.logError(ex);
-      logger.warn("Init Apollo Local Config failed - namespace: {}, reason: {}.",
-          m_namespace, ExceptionUtil.getDetailMessage(ex));
+      String logWarnMsg = "Init Apollo Local Config failed - namespace: {}, reason: {}.";
+      logger.warn(logWarnMsg, m_namespace, ExceptionUtil.getDetailMessage(ex));
+      DeferredLogUtil.warn(logger, logWarnMsg, m_namespace, ExceptionUtil.getDetailMessage(ex));
```
nobodyiam, line 64: Writing the `DeferredLogUtil.xxx` statements in many places don't seem like a very good way since it adds duplication and developer would probably forget to call this method in the future.

How about we wrap the original logger so that we don't need to duplicate the log statements? e.g.

```java
private static final Logger logger = DeferredLogger.wrap(LoggerFactory.getLogger(DefaultConfig.class));
```

or even

```java
private static final Logger logger = DeferredLoggerFactory.getLogger(DefaultConfig.class));
```

apollo-client/src/main/java/com/ctrip/framework/apollo/spring/boot/ApolloApplicationContextInitializer.java
```diff
@@ -91,7 +92,8 @@ public void initialize(ConfigurableApplicationContext context) {
   protected void initialize(ConfigurableEnvironment environment) {
 
     if (environment.getPropertySources().contains(PropertySourcesConstants.APOLLO_BOOTSTRAP_PROPERTY_SOURCE_NAME)) {
-      //already initialized
+      //already initialized ,  replay the logs that were printed before the logging system was initialized
+      DeferredLogUtil.replayTo();
```
nobodyiam, line 96: In a non-spring-boot environment, this code won't be executed, which means these logs remains in the memory forever?

I think we could do it in a more precise way, e.g.

**1. We only enable DeferredLogger when eager load is enabled**
Since `com.ctrip.framework.apollo.spring.boot.ApolloApplicationContextInitializer#postProcessEnvironment` will be the first lines of code executed, we could set a flag here

**2. In our DeferredLogger implementation, we cache the logs if the flag is true**

**3. We need to find a point to reset the flag to replay the cached logs and stop caching logs**
I assume there would be some sort of events published when the spring logging system is ready but we need to find it out.

klboke: - The log will not always be stored in the memory. In an environment where the replay condition cannot be triggered, the log will be recycled after one minute

- In the current client environment, I feel that whether it is the current design or the new delayed log design, it is possible to directly use the apollo.bootstrap.enabled tag to replay. The timing of the replay at this time is also more appropriate

apollo-client/src/main/java/com/ctrip/framework/apollo/internals/DefaultConfig.java
klboke: This is a good idea

klboke: After looking at it, it is not very easy to achieve. How about this?
```java
    public static void debug(Logger logger, String message, Object... objects) {
        logger.debug(message,objects);
        add(logger, Level.DEBUG, message, objects, null);
    }
```
Use it like this：
```java
      DeferredLogUtil.error(logger, "Initialize DefaultServerProvider failed.", ex);    
```

nobodyiam:
Here is my 2 cents, what do you think?
1. we keep the original logger interface but with a different implementation, e.g. a composite logger
2. we add a `deferred` flag to control the logger logic
3. in normal cases, the `deferred` flag is always false, so the log behavior is the same as the old one
4. in eager load case, we first set the `deferred` flag to true to cache the logs and after logging system is initialized, we set the flag to false and flush the cached logs

image

apollo-client/src/main/java/com/ctrip/framework/apollo/internals/DefaultConfig.java
```diff
@@ -59,7 +59,7 @@ private void initialize() {
     } catch (Throwable ex) {
       Tracer.logError(ex);
       logger.warn("Init Apollo Local Config failed - namespace: {}, reason: {}.",
-          m_namespace, ExceptionUtil.getDetailMessage(ex));
+              m_namespace, ExceptionUtil.getDetailMessage(ex));
```
nobodyiam, line 62: please use apollo code style to format the code

apollo-client/src/main/java/com/ctrip/framework/apollo/spring/boot/ApolloApplicationContextInitializer.java
```diff
@@ -91,7 +92,8 @@ public void initialize(ConfigurableApplicationContext context) {
   protected void initialize(ConfigurableEnvironment environment) {
 
     if (environment.getPropertySources().contains(PropertySourcesConstants.APOLLO_BOOTSTRAP_PROPERTY_SOURCE_NAME)) {
-      //already initialized
+      //already initialized ,  replay the logs that were printed before the logging system was initialized
```
nobodyiam, line 95:
```suggestion
      //already initialized, replay the logs that were printed before the logging system was initialized
```

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLog.java
```diff
@@ -0,0 +1,333 @@
...
+ * <pre class="code">
+ *  public void debug(String s, Object... objects);
+ *  public void info(String s, Object... objects);
+ *  public void warn(String s, Object... objects);
+ *  public void error(String s, Throwable throwable);
+ * </pre>
+ * @author kl (http://kailing.pub)
+ * @since 2021/5/19
+ */
+public class DeferredLog implements Logger {
```
nobodyiam, line 17:
```suggestion
public class DeferredLogger implements Logger {
```

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLog.java
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    @Override
+    public void warn(String s, Object o) {
+        logger.warn(s, o);
+    }
+
+    @Override
+    public void warn(String s, Object... objects) {
+        logger.warn(s, objects);
```
nobodyiam, line 229: I was thinking we add a check here so that we don't print the log twice? e.g.

```java
if (deferred) {
    DeferredLogCache.warn(logger, s, objects);
} else {
    logger.warn(s, objects);
}
```

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
```diff
@@ -0,0 +1,152 @@
...
+
+    private static void add(Logger logger, Level level, String message, Object[] objects, Throwable throwable) {
+        if (isEnabled()) {
+            Line logLine = new Line(level, message, objects, throwable, logger);
+            LOG_CACHE.put(LOG_INDEX.incrementAndGet(), logLine);
+        }
+    }
+
+    public static void replayTo() {
+        if (isEnabled()) {
```
nobodyiam, line 66: I think we need to disable deferred log first? or there might be some logs missing, i.e. those logs printed after the for statement

apollo-core/src/test/java/com/ctrip/framework/apollo/core/utils/DeferredLogCacheTest.java
```diff
@@ -0,0 +1,23 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author kl (http://kailing.pub)
+ * @since 2021/5/11
+ */
+public class DeferredLogCacheTest {
```
nobodyiam, line 12: This ut is a little bit simple...

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLog.java
nobodyiam: btw, I think we need to add some unit test for this class, or the code coverage rate will drop a lot..

nobodyiam:
@klboke please fix the failed tests

image

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Delayed log printing utility class, used only for logging when Apollo is initialized
+ *
+ * @author kl (http://kailing.pub)
+ * @since 2021/5/11
+ */
+public final class DeferredLogCache {
+
+  public static final int MAX_LOG_SIZE = 1000;
+  private static final AtomicBoolean ENABLED = new AtomicBoolean(false);
```
nobodyiam, line 21: How about we move this ENABLED property to DeferredLogger?

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Delayed log printing utility class, used only for logging when Apollo is initialized
+ *
+ * @author kl (http://kailing.pub)
+ * @since 2021/5/11
+ */
+public final class DeferredLogCache {
```
nobodyiam, line 18: I think we won't reuse this class elsewhere, how about we make it package-private or even move the whole class as a private inner static class of DeferredLogger?
```suggestion
final class DeferredLogCache {
```

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
```diff
@@ -0,0 +1,148 @@
...
+  private static final AtomicInteger LOG_INDEX = new AtomicInteger(0);
+  private static final Cache<Integer, Line> LOG_CACHE = CacheBuilder.newBuilder()
+      .maximumSize(MAX_LOG_SIZE)
+      .expireAfterWrite(1, TimeUnit.MINUTES)
+      .build();
+
+  private DeferredLogCache() {
+  }
+
+  public static synchronized void enableDeferredLog() {
```
nobodyiam, line 31: It looks like the `synchronized` is not necessary?

apollo-core/src/test/java/com/ctrip/framework/apollo/core/utils/DeferredLoggerTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @Test
+  public void testDeferredLog() {
+    DeferredLogCache.enableDeferredLog();
+
+    DeferredLoggerTest.logger.error("errorLogger");
+    DeferredLoggerTest.logger.info("inFoLogger");
+    DeferredLoggerTest.logger.warn("warnLogger");
+    DeferredLoggerTest.logger.debug("debugLogger");
+    DeferredLogCache.replayTo();
```
nobodyiam, line 60: I think we need to test the outContent is empty before replayTo method?

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
klboke: Yes, I forgot to remove it here

apollo-core/src/test/java/com/ctrip/framework/apollo/core/utils/DeferredLoggerTest.java
klboke: Yes it is

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
klboke: It doesn't feel very good, but here you can add a logic that only opens once, what do you think?

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
klboke: Same as above

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogger.java
```diff
@@ -0,0 +1,418 @@
...
+  public static void disable() {
+    state.set(0);
+  }
+
+  public static boolean isEnabled() {
+    return state.get() == 1;
+  }
+
+  public static void replayTo(){
+    DeferredLogCache.replayTo();
```
nobodyiam, line 50: How about we disable the state here so that DeferredLogCache is simply a cache which doesn't need to care whether the deferred logger is enabled or not?
```suggestion
    disable();
    DeferredLogCache.replayTo();
```

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
```diff
@@ -0,0 +1,140 @@
...
+        assert logLine != null;
+        Logger logger = logLine.getLogger();
+        Level level = logLine.getLevel();
+        String message = logLine.getMessage();
+        Object[] objects = logLine.getObjects();
+        Throwable throwable = logLine.getThrowable();
+        logTo(logger, level, message, objects, throwable);
+      }
+      clear();
+    }
```
nobodyiam, line 66: I think DeferredLogCache could remain simple by not caring whether DeferredLogger is enabled or not.
Actually it could even be renamed to `LogCache` which means it's simple a cache.

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogger.java
klboke: It makes sense, my thinking has been stuck on `DeferredLogUtils`

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
```diff
@@ -0,0 +1,140 @@
...
+
+  }
```
nobodyiam, line 68: I think DeferredLogCache could remain simple by not caring whether DeferredLogger is enabled or not.
Actually it could even be renamed to LogCache which means it's simple a cache.
```suggestion
  public static void replayTo() {
    for (int i = 1; i <= LOG_INDEX.get(); i++) {
      Line logLine = LOG_CACHE.getIfPresent(i);
      assert logLine != null;
      Logger logger = logLine.getLogger();
      Level level = logLine.getLevel();
      String message = logLine.getMessage();
      Object[] objects = logLine.getObjects();
      Throwable throwable = logLine.getThrowable();
      logTo(logger, level, message, objects, throwable);
    }
    clear();
  }
```

klboke: yes,it is

apollo-client/src/main/java/com/ctrip/framework/apollo/internals/DefaultConfig.java
```diff
@@ -29,7 +29,8 @@
  * @author Jason Song(song_s@ctrip.com)
  */
 public class DefaultConfig extends AbstractConfig implements RepositoryChangeListener {
-  private static final Logger logger = LoggerFactory.getLogger(DefaultConfig.class);
+
+  private static final Logger logger = DeferredLoggerFactory.getLogger(DefaultConfig.class);
```
nobodyiam, line 33: I think `com.ctrip.framework.apollo.internals.AbstractConfigFile` also need to use `DeferredLogger`

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogger.java
```diff
@@ -0,0 +1,419 @@
...
+    return logger.isWarnEnabled();
+  }
+
+  @Override
+  public void warn(String s) {
+    if (isEnabled()) {
+      DeferredLogCache.warn(logger, s);
+    } else {
+      logger.warn(s);
+    }
```
nobodyiam, line 287: Is it necessary to implement all the methods? maybe by calling the overloaded methods?

```suggestion
    this.warn(s, (Object) null);
```

apollo-client/src/main/java/com/ctrip/framework/apollo/internals/DefaultConfig.java
klboke: yes，It is

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogger.java
klboke: Basic coverage common methods

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
```diff
@@ -0,0 +1,152 @@
...
+
+  public static void debug(Logger logger, String message, Throwable throwable, Object... objects) {
+    add(logger, Level.DEBUG, message, objects, throwable);
+  }
+
+  public static void info(Logger logger, String message, Throwable throwable, Object... objects) {
+    add(logger, Level.INFO, message, objects, throwable);
+  }
+
+  public static void warn(Logger logger, String message, Throwable throwable, Object... objects) {
```
nobodyiam, line 38: with this signature, we can't make sure throwable and objects are not null at the same time, maybe we could add another overload method?

```java
public static void warn(Logger logger, String message, Object... objects) {}
public static void warn(Logger logger, String message, Throwable throwable) {}
```

nobodyiam: btw, looks like the ci failed, please take a look~

klboke: DeferredLogCache is not open to the public, the current api meets the needs, the compilation error is caused by my negligence

nobodyiam: OK, then I would suggest adding a method javadoc here to make it clear that user should not pass throwable and objects at the same time in case it was ignored in the future

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogger.java
```diff
@@ -0,0 +1,465 @@
...
+      DeferredLogCache.debug(logger, s, null, o);
+    } else {
+      logger.debug(s, o);
+    }
+  }
+
+  @Override
+  public void debug(String s, Object o, Object o1) {
+    if (isEnabled()) {
+      DeferredLogCache.debug(logger, s, null, o1);
```
nobodyiam, line 172: the object instance `o` is missing

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogger.java
```diff
@@ -0,0 +1,465 @@
...
+      DeferredLogCache.warn(logger, s, null, o, o1);
+    } else {
+      logger.warn(s, o, o1);
+    }
+  }
+
+  @Override
+  public void warn(String s, Throwable throwable) {
+    if (isEnabled()) {
+      DeferredLogCache.warn(logger, s, null);
```
nobodyiam, line 350: the throwable instance is missing

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
```diff
@@ -0,0 +1,168 @@
...
+      Level level = logLine.getLevel();
+      String message = logLine.getMessage();
+      Object[] objects = logLine.getObjects();
+      Throwable throwable = logLine.getThrowable();
+      logTo(logger, level, message, objects, throwable);
+    }
+    clear();
+  }
+
+  public static void clear() {
```
nobodyiam, line 82: this method could be package-private
```suggestion
 static void clear() {
```

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
```diff
@@ -0,0 +1,168 @@
...
+    LOG_CACHE.invalidateAll();
+    LOG_INDEX.set(0);
+  }
+
+  public static long logSize() {
```
nobodyiam, line 87: this method could be package-private
```suggestion
  static long logSize() {
```

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
```diff
@@ -0,0 +1,168 @@
...
+    return LOG_CACHE.size();
+  }
+
+  static void logTo(Logger logger, Level level, String message, Object[] objects,
```
nobodyiam, line 91: this method could be private
```suggestion
  private static void logTo(Logger logger, Level level, String message, Object[] objects,
```

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/DeferredLogCache.java
```diff
@@ -0,0 +1,168 @@
...
+        } else {
+          logger.error(message, objects);
+        }
+        break;
+      default:
+        throw new IllegalStateException("Unexpected value: " + level);
+    }
+  }
+
+  static class Line {
```
nobodyiam, line 127: this class could be private
```suggestion
  privarte static class Line {
```

apollo-client/src/main/java/com/ctrip/framework/apollo/internals/DefaultConfig.java
```diff
@@ -45,7 +45,8 @@
...
```
nobodyiam, line 49: Please also change LocalFileConfigRepository and RemoteConfigRepository to use `DeferredLogger`, I tested in my local environment and found the actual cause won't be printed in eager load mode

nobodyiam (approved):
LGTM

PR #3753, opened 2021-06-11
Fix typo on "load from specified cluster fist"

tianpeidong:
## Fix Typo

the comment should be "load from specified cluster first" instead of "load from specified cluster fist"


kezhenxu94 (approved):

Anilople (approved):
LGTM

PR #3755, opened 2021-06-12
Remove unused private field 'environment' of AutoUpdateConfigChangeListener.

mghio:

Remove unused private field 'environment' of AutoUpdateConfigChangeListener.

### DIFF
--- a/apollo-client/src/main/java/com/ctrip/framework/apollo/spring/property/AutoUpdateConfigChangeListener.java
+++ b/apollo-client/src/main/java/com/ctrip/framework/apollo/spring/property/AutoUpdateConfigChangeListener.java
@@ -39,18 +39,16 @@ public class AutoUpdateConfigChangeListener implements ConfigChangeListener{
   private static final Logger logger = LoggerFactory.getLogger(AutoUpdateConfigChangeListener.class);
 
   private final boolean typeConverterHasConvertIfNecessaryWithFieldParameter;
-  private final Environment environment;
   private final ConfigurableBeanFactory beanFactory;
   private final TypeConverter typeConverter;
   private final PlaceholderHelper placeholderHelper;
   private final SpringValueRegistry springValueRegistry;
   private final Gson gson;
 
-  public AutoUpdateConfigChangeListener(Environment environment, ConfigurableListableBeanFactory beanFactory){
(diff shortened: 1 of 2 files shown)


mghio:

apollo-client/src/main/java/com/ctrip/framework/apollo/spring/property/AutoUpdateConfigChangeListener.java
```diff
@@ -39,18 +39,16 @@
...
 
   private final boolean typeConverterHasConvertIfNecessaryWithFieldParameter;
-  private final Environment environment;
   private final ConfigurableBeanFactory beanFactory;
   private final TypeConverter typeConverter;
   private final PlaceholderHelper placeholderHelper;
   private final SpringValueRegistry springValueRegistry;
   private final Gson gson;
 
-  public AutoUpdateConfigChangeListener(Environment environment, ConfigurableListableBeanFactory beanFactory){
```
kezhenxu94, line 49: This is a public API and it resides in client module, so this is potentially a breaking change, though the users are rarely possible to use it, what do other reviewers' opinion ?

Anilople (commented):
apollo-client has been used widely by user.

compability is important.

If users extends those classes by themself, they will meet additionally trouble when method's signture changed.

Thanks for your time on this PR.

Reference:
* [语义化版本 2.0.0](https://semver.org/lang/zh-CN/)
* [Java开发手册（嵩山版）.pdf](https://github.com/alibaba/p3c/blob/master/Java%E5%BC%80%E5%8F%91%E6%89%8B%E5%86%8C%EF%BC%88%E5%B5%A9%E5%B1%B1%E7%89%88%EF%BC%89.pdf) page 45

apollo-client/src/main/java/com/ctrip/framework/apollo/spring/property/AutoUpdateConfigChangeListener.java
nobodyiam: I also agree we should be very very careful when introducing potentially breaking changes.
I checked the commit history and the environment is not used since #2173. But since this is a public class, I think it's better not to change the constructor signature.

kezhenxu94:
@mghio let's mark this and do the changes only when we must have breaking changes, for example, in the next major version. Thanks for the interest anyway 🙇 .

PR #3804, opened 2021-07-02
feature: add the delegating password encoder for apollo-portal simple auth

vdisk-group:

the best algorithm for password storage should change some day in the future. the `DelegatingPasswordEncoder` can makes the algorithm upgrade seamless.

## Brief changelog

1. extend the `ApolloPortalDB`.`Users`.`Password` to 512
2. replace the `BCryptPasswordEncoder` with the `DelegatingPasswordEncoder` backend of the `BCryptPasswordEncoder`
3. add a `PlaceholderPasswordEncoder` which return a random string as encoded password and never matches any encoded password.
4. replace the random string password for oidc with the `DelegatingPasswordEncoder` backend of the `PlaceholderPasswordEncoder`


nobodyiam:
This feature looks great!

However, I have one concern that if we need to change the password from "xxx" to "{bcrypt}xxx", does that mean the user could not upgrade to this version smoothly? e.g.

1. If the password is not changed, then the new version of apollo-portal won't work
2. If the password is changed, then the old version of apollo-portal won't work

So how about we leave the password in the original form and we let the new version of apollo-portal adapt them? Does the `org.springframework.security.crypto.password.DelegatingPasswordEncoder#setDefaultPasswordEncoderForMatches` work for this scenario?

vdisk-group:
> This feature looks great!
> 
> However, I have one concern that if we need to change the password from "xxx" to "{bcrypt}xxx", does that mean the user could not upgrade to this version smoothly? e.g.
> 
> 1. If the password is not changed, then the new version of apollo-portal won't work
> 2. If the password is changed, then the old version of apollo-portal won't work
> 
> So how about we leave the password in the original form and we let the new version of apollo-portal adapt them? Does the `org.springframework.security.crypto.password.DelegatingPasswordEncoder#setDefaultPasswordEncoderForMatches` work for this scenario?

add an adapter for old password.  
upgrade step:  
1. all of apollo-portal is old version, and the password has no prefix, all of apollo-portal works fine.
2. upgrade some of the apollo-portal to the new version, and the password has no prefix, the old version of the apollo-portal works fine as before and the new version of apollo-portal works fine with the help of the adapter.
3. upgrade all of apollo-portal, and the password has no prefix, the new version of apollo-portal works fine with the help of the adapter.
4. execute the sql scripts, change the password from "xxx" to "{bcrypt}xxx", all of the new version of apollo-portal works fine.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/spi/springsecurity/ApolloPasswordEncoderFactory.java
```diff
@@ -0,0 +1,76 @@
...
+        .put("sha256", new org.springframework.security.crypto.password.StandardPasswordEncoder());
+    encoders.put("argon2", new Argon2PasswordEncoder());
+
+    // placeholder encoder for oidc
+    encoders.put(PlaceholderPasswordEncoder.ENCODING_ID, new PlaceholderPasswordEncoder());
+    DelegatingPasswordEncoder delegatingPasswordEncoder = new DelegatingPasswordEncoder(encodingId,
+        encoders);
+
+    // todo: adapt the old password, and it should be removed in the next feature version of the 1.9.x
+    delegatingPasswordEncoder.setDefaultPasswordEncoderForMatches(new PasswordEncoderAdapter(encoders.get(encodingId)));
```
nobodyiam, line 73: Does it work if there is no PasswordEncoderAdapter?

```java
    delegatingPasswordEncoder.setDefaultPasswordEncoderForMatches(encoders.get(encodingId));
```

nobodyiam:
> execute the sql scripts, change the password from "xxx" to "{bcrypt}xxx", all of the new version of apollo-portal works fine.

Is this step mandatory? What if we just leave the password as in the form of "xxx"?

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/spi/springsecurity/ApolloPasswordEncoderFactory.java
vdisk-group: > Does it work if there is no PasswordEncoderAdapter?
> 
> ```java
> delegatingPasswordEncoder.setDefaultPasswordEncoderForMatches(encoders.get(encodingId));
> ```

it do works, but the `BCryptPasswordEncoder` can not throw an exact exception while the password has a wrong prefix just like `{foolish}xxx`. the `PasswordEncoderAdapter` is just used to detect the wrong prefix. see the `org.springframework.security.crypto.password.DelegatingPasswordEncoder.UnmappedIdPasswordEncoder`

vdisk-group:
> > execute the sql scripts, change the password from "xxx" to "{bcrypt}xxx", all of the new version of apollo-portal works fine.
> 
> Is this step mandatory? What if we just leave the password as in the form of "xxx"?

the smoothly upgrade feature provide by `DelegatingPasswordEncoder ` is dependent on the prefix.  
it do work now but once we upgrade the algorithm, the user who change the password after the upgrade can only login on the new version of apollo-portal.

nobodyiam:
> so it seems ok not to execute the sql scripts, but all of the apollo-portal need to update to the new version.

Right, this is a must and we should note it in the release note

vdisk-group:
> > execute the sql scripts, change the password from "xxx" to "{bcrypt}xxx", all of the new version of apollo-portal works fine.
> 
> Is this step mandatory? What if we just leave the password as in the form of "xxx"?

for better compatibility, we could split this pr into 2 parts.  
1. on the `1.9.0` add the `DelegatingPasswordEncoder` just as the decoder and just use the `BCryptPasswordEncoder` as the encoder as before. so the apollo-portal  of `1.9.0` can work well with the old version of apollo-portal.
2. on the next feature version, just like `1.10.0` replace the `BCryptPasswordEncoder` with `DelegatingPasswordEncoder` as the encoder. and then change the password from "xxx" to "{bcrypt}xxx". the apollo-portal  of `1.9.0` and `1.10.0` can both work well.


do you think it is necessary to do that?

nobodyiam:
I think it's not necessary to split this pr into 2 parts. It should be ok as long as the new version of apollo-portal could decode the password in both  `xxx` and `{bcrypt}xxx` forms so that we don't require users to do the data migration.
If the user changes the password during the apollo-portal upgrade process, then there are chances the user could not login (when accessing the old version apollo-portal) but I think this chance is trivial so it is ok.

apollo-client/src/test/java/com/ctrip/framework/apollo/spring/JavaConfigAnnotationTest.java
```diff
@@ -946,16 +947,16 @@ private void onChange(ConfigChangeEvent changeEvent) {
...
     @ApolloConfigChangeListener(value = SPECIAL_NAMESPACE, interestedKeyPrefixes = {"number",
         "logging.level"})
     private void onChange(ConfigChangeEvent changeEvent) {
       this.configChangeEventQueue.add(changeEvent);
     }
 
-    public ConfigChangeEvent getConfigChangeEvent() {
-      return this.configChangeEventQueue.poll();
+    public ConfigChangeEvent getConfigChangeEvent() throws InterruptedException {
+      return this.configChangeEventQueue.take();
```
nobodyiam, line 959: Is it better to use `E poll(long timeout, TimeUnit unit)` here?

apollo-portal/src/test/resources/sql/openapi/ConsumerServiceIntegrationTest.testFindAppIdsAuthorizedByConsumerId.sql
```diff
@@ -249,7 +249,7 @@ INSERT INTO `UserRole` (`Id`, `UserId`, `RoleId`, `DataChange_CreatedBy`, `DataC
 
 /*!40000 ALTER TABLE `Users` DISABLE KEYS */;
 INSERT INTO `Users` (`Id`, `Username`, `Password`, `UserDisplayName`, `Email`, `Enabled`) VALUES
-(1000, 'apollo', '$2a$10$7r20uS.BQ9uBpf3Baj3uQOZvMVvB1RN3PYoKE94gtz2.WAOuiiwXS', 'apollo', 'apollo@acme.com', 1);
+(1000, 'apollo', '{bcrypt}$2a$10$7r20uS.BQ9uBpf3Baj3uQOZvMVvB1RN3PYoKE94gtz2.WAOuiiwXS', 'apollo', 'apollo@acme.com', 1);
```
nobodyiam, line 252: shall we also remove the `{bcrypt}` here?

docs/zh/development/portal-how-to-implement-user-login-function.md
```diff
@@ -14,7 +14,7 @@ Apollo是配置管理系统，会提供权限管理（Authorization），理论
 
 |Id|Username|Password|Email|Enabled|
 |-|-|-|-|-|
-|1|apollo|$2a$10$7r20uS.BQ9uBpf3Baj3uQOZvMVvB1RN3PYoKE94gtz2.WAOuiiwXS|apollo@acme.com|1|
+|1|apollo|{bcrypt}$2a$10$7r20uS.BQ9uBpf3Baj3uQOZvMVvB1RN3PYoKE94gtz2.WAOuiiwXS|apollo@acme.com|1|
```
nobodyiam, line 17: I think we need to also remove the `{bcrypt}` here since if the user installs an old version of apollo-portal, the record in the database doesn't include `{bcrypt}`.

scripts/apollo-on-kubernetes/db/portal-db/apolloportaldb.sql
```diff
@@ -334,7 +334,7 @@ VALUES
 
 INSERT INTO `Users` (`Username`, `Password`, `UserDisplayName`, `Email`, `Enabled`)
 VALUES
-    ('apollo', '$2a$10$7r20uS.BQ9uBpf3Baj3uQOZvMVvB1RN3PYoKE94gtz2.WAOuiiwXS', 'apollo', 'apollo@acme.com', 1);
+    ('apollo', '{bcrypt}$2a$10$7r20uS.BQ9uBpf3Baj3uQOZvMVvB1RN3PYoKE94gtz2.WAOuiiwXS', 'apollo', 'apollo@acme.com', 1);
```
nobodyiam, line 337: I think we need to also remove the `{bcrypt}` here.

scripts/docker-quick-start/sql/apolloportaldb.sql
```diff
@@ -334,7 +334,7 @@ VALUES
 
 INSERT INTO `Users` (`Username`, `Password`, `UserDisplayName`, `Email`, `Enabled`)
 VALUES
-    ('apollo', '$2a$10$7r20uS.BQ9uBpf3Baj3uQOZvMVvB1RN3PYoKE94gtz2.WAOuiiwXS', 'apollo', 'apollo@acme.com', 1);
+    ('apollo', '{bcrypt}$2a$10$7r20uS.BQ9uBpf3Baj3uQOZvMVvB1RN3PYoKE94gtz2.WAOuiiwXS', 'apollo', 'apollo@acme.com', 1);
```
nobodyiam, line 337: I think we need to also remove the `{bcrypt}` here.

scripts/sql/apolloportaldb.sql
```diff
@@ -335,7 +335,7 @@ VALUES
 
 INSERT INTO `Users` (`Username`, `Password`, `UserDisplayName`, `Email`, `Enabled`)
 VALUES
-	('apollo', '$2a$10$7r20uS.BQ9uBpf3Baj3uQOZvMVvB1RN3PYoKE94gtz2.WAOuiiwXS', 'apollo', 'apollo@acme.com', 1);
+	('apollo', '{bcrypt}$2a$10$7r20uS.BQ9uBpf3Baj3uQOZvMVvB1RN3PYoKE94gtz2.WAOuiiwXS', 'apollo', 'apollo@acme.com', 1);
```
nobodyiam, line 338: shall we also remove the {bcrypt} here?

nobodyiam (approved):
LGTM

PR #3925, opened 2021-08-27
change scrollbar css

zeemood:

Make scrollbar visible and beautiful. For the scene of hundreds config items, it seems like that a scrollbar takes me more convenience because of its high scrollview 

Fixes #

## Brief changelog

change scrollbar css

- [X] Read the [Contributing Guide](https://github.com/ctripcorp/apollo/blob/master/CONTRIBUTING.md) before making this pull request.
- [X] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [X] Write necessary unit tests to verify the code.
- [X] Run `mvn clean test` to make sure this pull request doesn't break anything.
- [X] Update the [`CHANGES` log](https://github.com/ctripcorp/apollo/blob/master/CHANGES.md).


nobodyiam (changes requested):
This feature looks great! Would you please also update the CHANGES.md?

apollo-portal/src/main/resources/static/styles/common-style.css
```diff
@@ -534,18 +534,19 @@ table th {
 
 /*搜索框*/
 ::-webkit-scrollbar {
```
nobodyiam, line 536: It seems the best way is to remove the -webkit-scrollbar related styles?

zeemood: yep. it can keep the same style of the scrollbar in different browser.

nobodyiam: There is a similar pr #3936 which simply removes the -webkit-scrollbar related styles and I think it also works well. We may need to decide which way to go. @lepdou do you have any suggestion?

zeemood: [#3936](https://github.com/apolloconfig/apollo/pull/3936) that's my colleague.  It's good enough for me.

PR #3937, opened 2021-09-01
Remove spring dependencies from internal code

klboke:
The internal core code cannot couple Spring's dependencies, which results in limited functionality in pure Java API mode


JaredTan95 (approved):
LGTM

PR #3943, opened 2021-09-02
refactor: let open api more easier to use and development

### LINKED ISSUE #3942: openapi design. client-server mode. let controller to implement interface

If someone want to find which controller in portal implement the method `com.ctrip.framework.apollo.openapi.client.ApolloOpenApiClient#createCluster`, it will take some effort.

How about add some interface to make it more easier?

For example:

* interface `ApolloXXXOpenApi`: in module `apollo-openapi`, user can use it directly.
* class `ApolloXXXOpenApiClient`: in module `apollo-openapi`, the client's implementation of `ApolloXXXOpenApi`, user cannot access it.
* class `ApolloXXXOpenApiController`: in module `apollo-portal`, the server's implementation of `ApolloXXXOpenApi`

Use this pattern, we can add more openapi faster than before?

**Describe the solution you'd like**
Keep the compability, add interfaces to module `apollo-openapi`, let controllers in module `apollo-portal` implement them.

**Describe alternatives you've considered**
It is ok to use `spring-web` in module `apollo-openapi`? We can maintain the url information in ApolloXXXOpenApi

```java
public interface ApolloNamespaceOpenApi {
  @GetMapping(value = "/openapi/v1/envs/{env}/apps/{appId}/clusters/{clusterName}/namespaces")
  List<OpenNamespaceDTO> findNamespaces(String appId, String env, String clusterName);
}
```
Then the client and server can use this information, the user also can view it directly.

Drawback: `apollo-openapi` is a lower level than user's code, so `apollo-openapi` cannot use spring cause it maybe conflict with user's code

Anilople:

Let open api more easier to use and development.

Just a simple demo, not the final code.

Fixes #3942

## Brief changelog

Add interfaces ApolloXXXOpenApi under java package `com.ctrip.framework.apollo.openapi.api`.

Let controller under java package `com.ctrip.framework.apollo.openapi.v1.controller` implement each ApolloXXXOpenApi as server.

An interface ApolloOpenApi extends them all, let com.ctrip.framework.apollo.openapi.client.ApolloOpenApiClient implement ApolloOpenApi with compability.


apollo-openapi/src/main/java/com/ctrip/framework/apollo/openapi/client/ApolloOpenApiClient.java
```diff
@@ -115,36 +121,54 @@ private ApolloOpenApiClient(String portalUrl, String token, RequestConfig reques
...
   /**
-   * Create the app namespace
+   * Create the app namespace. Suggest use {@link #createNamespace(String, OpenAppNamespaceDTO)} instead.
    */
   public OpenAppNamespaceDTO createAppNamespace(OpenAppNamespaceDTO appNamespaceDTO) {
     return namespaceService.createAppNamespace(appNamespaceDTO);
   }
 
+  @Override
+  public OpenAppNamespaceDTO createNamespace(String appId, OpenAppNamespaceDTO appNamespaceDTO) {
```
nobodyiam, line 164: How about keeping the original method name?
```suggestion
  public OpenAppNamespaceDTO createAppNamespace(String appId, OpenAppNamespaceDTO appNamespaceDTO) {
```

apollo-portal/src/main/java/com/ctrip/framework/apollo/openapi/v1/controller/ClusterController.java
```diff
@@ -60,7 +61,7 @@ public OpenClusterDTO loadCluster(@PathVariable("appId") String appId, @PathVari
   @PreAuthorize(value = "@consumerPermissionValidator.hasCreateClusterPermission(#request, #appId)")
   @PostMapping(value = "apps/{appId}/clusters")
   public OpenClusterDTO createCluster(@PathVariable String appId, @PathVariable String env,
-      @Valid @RequestBody OpenClusterDTO cluster, HttpServletRequest request) {
```
nobodyiam, line 63: This is a very creative idea, by implementing the same interface, it's easier to write the implementation in both client and server side.
However, the downside is this would require the server side's signature the same as the client side, e.g. we have to delete the `HttpServletRequest request` here. That would have some impact on the server side implementation, e.g. I think the `@PreAuthorize(value = "@consumerPermissionValidator.hasCreateClusterPermission(#request, #appId)")` would not work now?

Anilople: How about use an `@Override` method always `throw new UnsupportedOperationException();` to satisfy the type checking, 
another method with HttpServletRequest for the `@PreAuthorize`, 
and add some javadoc to illustrate it?

For example, in class `com.ctrip.framework.apollo.openapi.v1.controller.ClusterController`
```java
  @PreAuthorize(value = "@consumerPermissionValidator.hasCreateClusterPermission(#request, #appId)")
  @PostMapping(value = "apps/{appId}/clusters")
  public OpenClusterDTO createCluster(@PathVariable String appId, @PathVariable String env,
      @Valid @RequestBody OpenClusterDTO cluster, HttpServletRequest request) {
    // ...
  }

  /**
   * The method invoked really is {@link #createCluster(String, String, OpenClusterDTO, HttpServletRequest)} for permission check.
   */
  @Override
  public OpenClusterDTO createCluster(String appId, String env, OpenClusterDTO openClusterDTO) {
    throw new UnsupportedOperationException();
  }
```

There are 2 method `createCluster`, one with `@Override` for type checking, another one with `HttpServletRequest` for `@PreAuthorize`, and we add some javadoc to illustrate it.

nobodyiam: In this case, I think it doesn't make much sense to define the interfaces as they are only implemented in client side.
One possible way is to let createCluster with HttpServletRequest param to call the createCluster method, but this seems quite bloated and not elegant.

Anilople: Can we reverse the invokation relationship? 

i.e keep createCluster with HttpServletRequest only for permission validation and url mapping, 
then call createCluster method without HttpServletRequest  param.
 
```java
  @PreAuthorize(value = "@consumerPermissionValidator.hasCreateClusterPermission(#request, #appId)")
  @PostMapping(value = "apps/{appId}/clusters")
  public OpenClusterDTO createCluster(@PathVariable String appId, @PathVariable String env,
      @Valid @RequestBody OpenClusterDTO cluster, HttpServletRequest request) {
    return this.createCluster(appId, env, openClusterDTO);
  }
  @Override
  public OpenClusterDTO createCluster(String appId, String env, OpenClusterDTO openClusterDTO) {
    // ...
  }
```

nobodyiam: This is what I mentioned in the previous comment. The down side is it makes the code quite bloated and not elegant..

Anilople: Maybe we can get HttpServletRequest by spring expression without push it to method parameter.

The origin java code is 

```java
HttpServletRequest httpServletRequest = ((ServletRequestAttributes) RequestContextHolder.currentRequestAttributes()).getRequest();
```

We can change to spring expression

```
T(org.springframework.web.context.request.RequestContextHolder).currentRequestAttributes().getRequest()
```

then use it to instead of `#request`?

Reference:
* https://docs.spring.io/spring-framework/docs/4.3.10.RELEASE/spring-framework-reference/html/expressions.html#expressions-types
* https://stackoverflow.com/questions/7585627/calling-static-methods-from-spring-security-expressions/7585812

nobodyiam: How about we define the interfaces and implement them both in client and server side as dedicated services?
e.g. 
1. define interface ClusterOpenApiService
2. implement it in apollo-openapi as ClientClusterOpenApiService which will be invoked by ApolloOpenApiClient
3. implement it in apollo-portal as ServerClusterOpenApiService which will be invoked by ClusterController
4. ClusterController will preserve the `PreAuthorize` annotation but most of the logic will be removed to ServerClusterOpenApiService

Anilople: There is `com.ctrip.framework.apollo.openapi.client.service.ClusterOpenApiService` already.

If 
> implement it in apollo-openapi as ClientClusterOpenApiService which will be invoked by ApolloOpenApiClient

It should rename to `ClientClusterOpenApiService`.

Should we keep the class name for compability? Just let it implement interface `com.ctrip.framework.apollo.openapi.api.ClusterOpenApiService`. i.e class name as same as interface's

nobodyiam: I think we could leave the name as it is now for compability consideration :-)

nobodyiam (changes requested):
Thanks for the contribution! It looks good to me, would you please update the CHANGES.md?

Anilople:
> Thanks for the contribution! It looks good to me, would you please update the CHANGES.md?

Done

nobodyiam (approved):
LGTM

PR #3983, opened 2021-09-14
fix NullPointerException hazard in StringUtils.join(..) method

WillardHu:
Signed-off-by: WillardHu <wei.hu@daocloud.io>

When the `collection` is null, a `NullPointerException` is thrown here. Otherwise `iterator == null` is always false.

Fixes #

## Brief changelog

fix NullPointerException hazard in StringUtils.join(..) method.

### DIFF
--- a/apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/StringUtils.java
+++ b/apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/StringUtils.java
@@ -344,44 +344,4 @@ public class StringUtils {
     String format(T obj);
   }
 
-  public static <T> String join(Collection<T> collection, String separator) {
-    return join(collection, separator, new StringFormatter<T>() {
-      @Override
-      public String format(T obj) {
-        return obj.toString();
-      }
-    });
-  }
-
-  public static <T> String join(Collection<T> collection, String separator,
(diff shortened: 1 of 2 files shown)


apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/StringUtils.java
```diff
@@ -355,11 +355,11 @@ public String format(T obj) {
 
   public static <T> String join(Collection<T> collection, String separator,
```
kezhenxu94, line 356: Seems this method is only used in tests, let's just remove it

kezhenxu94 (approved):
LGTM, although this is kind of breaking changes in `apollo-client`, I think these 2 methods are not meant to be used by users, and it's easy to migrate to other string util, I'm OK to remove them, WDYT @nobodyiam

apollo-core/src/main/java/com/ctrip/framework/apollo/core/utils/StringUtils.java
WillardHu: Removed method and changed the commit message to 'delete unused method StringUtils.join(..)'

nobodyiam:
@kezhenxu94 I agree with you that it's ok to remove these 2 methods

Issue #3995, opened 2021-09-20
The WebHook feature doesn't include the changed configurations


nisiyong:
Recently I wanna [enable the webhook notification](https://www.apolloconfig.com/#/zh/development/portal-how-to-enable-webhook-notification) to do something on an external application. 

But this feature can't meet me cause it doesn't include the changes configurations when releasing namespace. Currently, this feature ONLY contains all the newest items in the `configuration` field.

What about adding the changed data in this webhook data likes the [Email notification](https://www.apolloconfig.com/#/zh/development/portal-how-to-enable-email-service?id=_32-emailtemplatereleasemodulediff)? In my opinion,  the difference between the two releases is more meaningful.

```
{
    "appId": "",  // appId
    "clusterName": "",  // 集群
    "namespaceName": "", // namespace
    "operator": "",  // 发布人
    "releaseId": 2,  // releaseId
    "releaseTitle": "",  // releaseTitle 
    "releaseComment": "",  // releaseComment
    "releaseTime": "",  // 发布时间  eg：2020-01-01T00:00:00.000+0800
    "configuration": [ { // 发布后的全部配置，如果为灰度发布，则为灰度发布后的全部配置
        "firstEntity": "",  // 配置的key
        "secondEntity": ""  // 配置的value
    } ],
    "isReleaseAbandoned": false,
    "previousReleaseId": 1,  // 上一次正式发布的releaseId
    "operation":  // 0-正常发布 1-配置回滚 2-灰度发布 4-全量发布
    "operationContext": {  // 操作设置的属性配置
        "isEmergencyPublish": true/false,  // 是否紧急发布
        "rules": [ {  // 灰度规则
            "clientAppId": "",   // appId
            "clientIpList": [ "10.0.0.2", "10.0.0.3" ]  // IP列表
        } ],
        "branchReleaseKeys": [ "", "" ]  // 灰度发布的key
    }
}
```

nobodyiam:
What's the actual use case? There is a previousReleaseId in the webhook request body, which could be used to retrieve the previous configuration.

nisiyong:
Webhook is good for users to do something customize, like notification bot, etc.

Even the email feature, I think if the webhook data support shows the diff between the two releases, users could integrate their message system with RESTful.

nobodyiam:
Your point makes sense to me. The wehook, email feature are all async notifications triggered when the configuration is released. Actually they are implemented in the same program block so I think it's reasonable to add difference between the two releases.

[...]

PR #3998, opened 2021-09-23
refactor: unify PropertyChangeType

nisiyong:

unify PropertyChangeType, prepare for #3995 

## Brief changelog

move PropertyChangeType to the core package.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@ Apollo 1.10.0
 * [refactor: let open api more easier to use and development](https://github.com/apolloconfig/apollo/pull/3943)
 * [feat(scripts): use bash to call openapi](https://github.com/apolloconfig/apollo/pull/3980)
 * [Support search by item](https://github.com/apolloconfig/apollo/pull/3977)
+* [refactor: Unify PropertyChangeType](https://github.com/apolloconfig/apollo/pull/3998)
 
 ------------------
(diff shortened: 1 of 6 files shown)


Anilople:
User may use class `com.ctrip.framework.apollo.model.ConfigChange` widely in apollo-client already. 

Consider code like follow
```java
    @ApolloConfigChangeListener
    private void onChange(ConfigChangeEvent changeEvent) {
      ConfigChange configChange = changeEvent.getChange("a");
      if (PropertyChangeType.ADDED.equals(configChange.getChangeType())) {
        // do something
      }
    }
```

Get ConfigChange from event, get PropertyChangeType from ConfigChange.

If we tell user should use enum `com.ctrip.framework.apollo.core.enums.PropertyChangeType`, the `equals` will always return `false` because 2 enum's type are mismatch.

i.e 
```java
      if (com.ctrip.framework.apollo.core.enums.PropertyChangeType.ADDED.equals(configChange.getChangeType())) {
        // do something
      }
```
always return `false`.

To catch correct PropertyChangeType, user still have to use `com.ctrip.framework.apollo.enums.PropertyChangeType` which in module `apllo-client`?

nobodyiam:
The `com.ctrip.framework.apollo.enums.PropertyChangeType` has been used widely so I suggest we keep it even though it's better to move it to apollo-core module.
With that said, why shall we unify the PropertyChangeType, or what's the problem if we don't unify it?

PR #4082, opened 2021-11-09
Fix flaky test testGetPropertyNames.

yyfMichaelYan:
Fix the flaky test testGetPropertyNames due to the non-deterministic iteration order of HashSet.

This flaky test is found by using Nondex, a tool used to find flaky tests in Java projects. The original assertion on Line#69 may fail because the iteration order of HashSet is non-deterministic, and the elements in the array converted from ```somePropertyNames``` may not be the same as the array ```result```. By converting ```result``` to a HashSet, the new assertion ```assertEquals``` will only check the elements are equal in both sets so that the test will always pass.


yyfMichaelYan:

nobodyiam (approved):
Thanks!

PR #4083, opened 2021-11-09
Fix flaky test testAssembleQueryConfigUrl.

yyfMichaelYan:
Fix flaky test testAssembleQueryConfigUrl due to the non-deterministic iteration order of HashMap.

This flaky test is found by using Nondex, a tool used to find flaky tests in Java projects. The assertion on Line#361 in ```RemoteConfigRepositoryTest.java``` may fail because the order of two entries, ```someKey``` and ```anotherKey```, is non-deterministic, so the converted string of ```queryConfigUrl``` is non-deterministic, resulting in the flaky test. To fix this test, the data structure of the ```details``` field of ```ApolloNotificationMessages``` is changed from ```HashMap``` to ```TreeMap``` so that the entries of the map is sorted. As a result, the iteration order is deterministic and the test will always pass.

### DIFF
--- a/apollo-core/src/main/java/com/ctrip/framework/apollo/core/dto/ApolloNotificationMessages.java
+++ b/apollo-core/src/main/java/com/ctrip/framework/apollo/core/dto/ApolloNotificationMessages.java
@@ -28,7 +28,7 @@ public class ApolloNotificationMessages {
   private Map<String, Long> details;
 
   public ApolloNotificationMessages() {
-    this(Maps.<String, Long>newHashMap());
+    this(Maps.<String, Long>newTreeMap());
   }
 
   private ApolloNotificationMessages(Map<String, Long> details) {


nobodyiam:
Thanks for the report. In this case, though the converted string of queryConfigUrl is non-deterministic due to hash map characteristic, I think this test will always pass as the actual logic to handle `notificationMessages` in `assembleQueryConfigUrl` is the same as `testAssembleQueryConfigUrl`.

assembleQueryConfigUrl:
```java
    if (remoteMessages != null) {
      queryParams.put("messages", queryParamEscaper.escape(GSON.toJson(remoteMessages)));
    }
```

testAssembleQueryConfigUrl:
```java
    assertTrue(queryConfigUrl
        .contains("messages=" + UrlEscapers.urlFormParameterEscaper()
            .escape(gson.toJson(notificationMessages))));
```

yyfMichaelYan:
Thanks for your response! Yes, both ```assembleQueryConfigUrl``` and ```testAssembleQueryConfigUrl``` use the same method ```toJson``` to convert the Object to JSON. However, the returned JSON may not always be the same every time ```toJson``` is called. If you check the implementation of ```toJson```, you will find in the end that it needs to iterate the HashMap to convert it to JSON, and the iteration order is non-deterministic. In other words, it's possible that ```entryA``` comes before ```entryB``` the first time ```toJson``` is called, and ```entryB``` comes before ```entryA``` the second time ```toJson``` is called.

nobodyiam (approved):
LGTM

PR #4185, opened 2021-12-31
Allow to run unit tests on jdk17

Shoothzj:
Allow to run unit tests on jdk17

## Brief changelog
add jdk17 profile, add test arguments


Shoothzj:
@nobodyiam PTAL, thanks :)

pom.xml
```diff
@@ -730,6 +730,15 @@
 				</plugins>
 			</build>
 		</profile>
+		<profile>
+			<id>jdk17</id>
+			<activation>
+				<jdk>[17,)</jdk>
+			</activation>
+			<properties>
+				<test.additional.args> --add-opens java.base/java.lang=ALL-UNNAMED </test.additional.args>
```
nobodyiam, line 739: What does this line mean?

nobodyiam (commented):
Thanks for raising this pr. Currently apollo unit tests could only run well with JDK 8, so it would be great if it could run with JDK 11 and JDK 17.
However, I'm not sure what this pr could achieve, does it intend to make the ut work on JDK 17?

Shoothzj:
> Thanks for raising this pr. Currently apollo unit tests could only run well with JDK 8, so it would be great if it could run with JDK 11 and JDK 17.
> 
> However, I'm not sure what this pr could achieve, does it intend to make the ut work on JDK 17?

@nobodyiam This pr allows people run unit test with jdk17,but there are still some tests can not pass, specially client module, we need to bump springboot from 1.5.16, it's a bigger work so I suggest split

pom.xml
Shoothzj: This line means opens `java.lang` module in java.base

`apollo` unit test code is development under java8，adding this line to avoid `module java.base does not "opens java.lang" to unnamed module @224aed64` error.

nobodyiam:
@Shoothzj 

Thanks for the clarification. Maybe we could do it in a reverse order, e.g. first solve the prerequisites like `bump springboot from 1.5.16`, then we submit the pr to make the unit tests work on jdk 17?

PR #4198, opened 2022-01-07
The release history of namespaces that are not properties will also show comments and release times

klboke:

- The release history of namespaces that are not properties will also show comments and release times

- It is important to keep the UI experience consistent with properties, and then the precise release time. Here is a comparison of the before and after changes.

image
After the change
image


nobodyiam (approved):
Looks great!

PR #4213, opened 2022-01-20
optimize create namespace page

lepdou:

optimize create namespace page

## Brief changelog
change list

1. adjust form field order
2. make create namespace as default action
3. make create private namespace as default type

#### old page
image

#### new page
image


apollo-portal/src/main/resources/static/scripts/controller/NamespaceController.js
```diff
@@ -66,7 +66,7 @@ namespace_module.controller("LinkNamespaceController",
                 appId: $scope.appId,
                 name: '',
                 comment: '',
-                isPublic: true,
+                isPublic: false,
```
nobodyiam, line 69: > make create private namespace as default type

Will this change bring confusion to the user as the default behavior is changed?

lepdou: Yes，I think the most scene is creating private namespace not public namespace. 

I changed select namespace type field above input namespace name field. So the user clearly knows the type of namespace being created.

nobodyiam:
I'm okay with this change then.
BTW, does it look better if we align the create namespace tab with the appid?

image

image

PR #4260, opened 2022-03-10
add self define config

### LINKED ISSUE #4259: feat: add self define meta server
In some environment，we put the config service and the admin service under the gateway, But for the portal we want to access the config service and admin service with gateway not the host which admin service and config service in, Such as different k8s environment. 
image

 So we want flexible config the admin service and the config service, and remove the stateful of config service(eureka).

gy09535:
Fixes #4259 

## Brief changelog


gy09535:

gy09535:
recheck

nobodyiam:
It looks like self define discovery is very similar to [KubernetesDiscoveryService](https://github.com/apolloconfig/apollo/blob/master/apollo-configservice/src/main/java/com/ctrip/framework/apollo/metaservice/service/KubernetesDiscoveryService.java), could we just reuse that instead?

gy09535:
> 
Functionally speaking, it is similar, but semantically it is different, k8s definition is through k8s service discovery to replace eureka, but in some non-k8s environments or complex deployment scenarios, we do not need to use k8s services as service discovery, then we need more flexible semantics to define this operation, if we use k8s service discovery, will not cause semantic mixing, conceptually unclear.

gy09535:
> It looks like self define discovery is very similar to [KubernetesDiscoveryService](https://github.com/apolloconfig/apollo/blob/master/apollo-configservice/src/main/java/com/ctrip/framework/apollo/metaservice/service/KubernetesDiscoveryService.java), could we just reuse that instead?

Functionally speaking, it is similar, but semantically it is different, k8s definition is through k8s service discovery to replace eureka, but in some non-k8s environments or complex deployment scenarios, we do not need to use k8s services as service discovery, then we need more flexible semantics to define this operation, if we use k8s service discovery, can cause semantic mixing, conceptually unclear.

nobodyiam:
Then I suppose we could do like this, in which `custom-defined` is the profile to be activated for those non-kubernetes environment, but we could share the same implementation.

```java
@Service
@Profile({"kubernetes", "custom-defined"})
public class KubernetesDiscoveryService implements DiscoveryService {
}
```

gy09535:
> Then I suppose we could do like this, in which `custom-defined` is the profile to be activated for those non-kubernetes environment, but we could share the same implementation.
> 
> ```java
> @Service
> @Profile({"kubernetes", "custom-defined-discovery"})
> public class KubernetesDiscoveryService implements DiscoveryService {
> }
> ```

good idea 👍

PR #4314, opened 2022-04-15
Update ReleaseMessageServiceWithCacheTest.java

plzdoo:
Test `testWhenReleaseMsgSizeBiggerThan500` in class `ReleaseMessageServiceWithCacheTest` is flaky due to the iteration of hashset used in method `findLatestReleaseMessagesGroupByMessages`  in class `ReleaseMessageServiceWithCache`. It may occasionally fail in different JVM's or in different Java versions.

I modify the test so this test won't be affect by the flakiness of indeterministic order of iteration of hashset.

reproduce:
`mvn -pl apollo-configservice edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=ReleaseMessageServiceWithCacheTest#testWhenReleaseMsgSizeBiggerThan500 -DnondexRuns=10`

### DIFF
--- a/apollo-configservice/src/test/java/com/ctrip/framework/apollo/configservice/service/ReleaseMessageServiceWithCacheTest.java
+++ b/apollo-configservice/src/test/java/com/ctrip/framework/apollo/configservice/service/ReleaseMessageServiceWithCacheTest.java
@@ -22,6 +22,7 @@ import com.ctrip.framework.apollo.biz.message.Topics;
 import com.ctrip.framework.apollo.biz.repository.ReleaseMessageRepository;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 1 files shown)


apollo-configservice/src/test/java/com/ctrip/framework/apollo/configservice/service/ReleaseMessageServiceWithCacheTest.java
```diff
@@ -156,8 +157,8 @@ public void testWhenReleaseMsgSizeBiggerThan500() throws Exception {
             .findLatestReleaseMessagesGroupByMessages(Sets.newHashSet(someMsgContent, antherMsgContent));
```
nobodyiam, line 157: Good point, how about we change the input to `LinkedHashSet`? That might look more elegant.

plzdoo: Thanks for reply.
Yes it will also works. It just depends on whether you want the assertion to be depend on order or not. I have modified the fix to use `LinkedHashSet`.

nobodyiam (approved):
LGTM

PR #4423, opened 2022-06-17
Fix apolloconfigdb-v190-v200.sql execution error

klboke:

```sql
ALTER TABLE `Item`
    ADD COLUMN `DeletedAt` BIGINT(20) NOT NULL DEFAULT '0' COMMENT 'Delete timestamp based on milliseconds' AFTER `IsDeleted`,
    ADD INDEX IX_key (`Key`);
```
 the IX_key index already exists in the old version. The execution times an error.
`Duplicate key name 'IX_key'`


nobodyiam:
I checked the schema of [1.9.0](https://github.com/apolloconfig/apollo/blob/v1.9.0/scripts/sql/apolloconfigdb.sql#L225) and there is no IX_key?

klboke:
I re-checked, it is indeed the problem of the previous upgrade action on my side, I will close this pull request

PR #4425, opened 2022-06-21
Upgrade mysql-connector-java version to fix possible transaction rollback failure issue

klboke:

There is a known bug in mysql-connector-java:8.0.28 that prevents transaction rollback when useLocalSessionState=true. apollo does not disallow users to set mysql link parameters, so there may be a potential problem

- This bug was fixed in mysql-connector-java:8.0.29

-   Reference: https://dev.mysql.com/doc/relnotes/connector-j/8.0/en/news-8-0-29.html
image

- Bug mechanism reference：https://my.oschina.net/klblog/blog/5542934


nobodyiam (approved):
LGTM

Shoothzj:
@nobodyiam `mysql-connector-java` is under `GPL 2.0` license. Can we use it directly?

klboke:
I don't know much about the protocol rules, but I see that spring-boot and skywalking also have this dependency, and they are both open source projects for the Apache protocol. So, it shouldn't be a big problem. No other questions then, I'll merge it in first

PR #4428, opened 2022-06-27
Optimize Spring-Security Firewall Deny Request Response 400

klboke:
- Before modification.
Requesting a service using a non-compliant URL specification returns 500 and the server side reports an exception, e.g.
image

- After modification.
Return 400 , e.g.
image

This scenario is more appropriate to use Bad Request (400), in our production environment, because of this problem, the gateway layer alerts often misreported Apollo Server exceptions


nobodyiam (approved):
LGTM

PR #4435, opened 2022-07-01
refactor:Simplify the code

zhangyangx:
Signed-off-by: yang.zhang <yang.zhang-iog@daocloud.io>

Simplify the code

Fixes #

### DIFF
--- a/apollo-core/src/main/java/com/ctrip/framework/foundation/internals/NetworkInterfaceManager.java
+++ b/apollo-core/src/main/java/com/ctrip/framework/foundation/internals/NetworkInterfaceManager.java
@@ -119,15 +119,13 @@ public enum NetworkInterfaceManager {
 
     try {
       Enumeration<NetworkInterface> interfaces = NetworkInterface.getNetworkInterfaces();
-      List<NetworkInterface> nis = interfaces == null ? Collections.emptyList()
-          : Collections.list(interfaces);
+      if (interfaces == null) {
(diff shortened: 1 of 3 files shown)


apollo-core/src/main/java/com/ctrip/framework/foundation/internals/NetworkInterfaceManager.java
```diff
@@ -119,15 +119,13 @@ private void load() {
 
     try {
       Enumeration<NetworkInterface> interfaces = NetworkInterface.getNetworkInterfaces();
-      List<NetworkInterface> nis = interfaces == null ? Collections.emptyList()
-          : Collections.list(interfaces);
+      if (interfaces == null) {
+        m_local = InetAddress.getLoopbackAddress();
+        return;
+      }
+      List<NetworkInterface> nis = Collections.list(interfaces);
```
nobodyiam, line 126: After reading the logic of this method, I think this by-pass logic is correct. But I'm wondering how about we simply leave the original logic as it is?
```java
List<NetworkInterface> nis = interfaces == null ? Collections.emptyList()
          : Collections.list(interfaces);
```

zhangyangx: The original code is fine, the new code should be fine, and the updated code performs better

nobodyiam (approved):
LGTM

PR #4451, opened 2022-07-08
Unify the length of the NamespaceName field in all tables to 500

klboke:

Unify the length of the NamespaceName field in all tables to 500

- Before the modification, some NamespaceNames are varchar(32) and some are varchar(500) . The App's Name is already varchar(500) , so uniformly adjust the NamespaceName to 500.


scripts/docker-quick-start/sql/apolloconfigdb.sql
```diff
@@ -161,9 +161,9 @@ CREATE TABLE `GrayReleaseRule` (
   `Id` int(11) unsigned NOT NULL AUTO_INCREMENT COMMENT '主键',
   `AppId` varchar(64) NOT NULL DEFAULT 'default' COMMENT 'AppID',
   `ClusterName` varchar(32) NOT NULL DEFAULT 'default' COMMENT 'Cluster Name',
-  `NamespaceName` varchar(32) NOT NULL DEFAULT 'default' COMMENT 'Namespace Name',
+  `NamespaceName` varchar(500) NOT NULL DEFAULT 'default' COMMENT 'Namespace Name',
   `BranchName` varchar(32) NOT NULL DEFAULT 'default' COMMENT 'branch name',
-  `Rules` varchar(16000) DEFAULT '[]' COMMENT '灰度规则',
+  `Rules` longtext DEFAULT '[]' COMMENT '灰度规则',
```
nobodyiam, line 166: I think maybe `text`(with length 65535) is long enough?
```suggestion
  `Rules` text DEFAULT '[]' COMMENT '灰度规则',
```

scripts/docker-quick-start/sql/apolloconfigdb.sql
```diff
@@ -61,7 +61,7 @@ DROP TABLE IF EXISTS `AppNamespace`;
 
 CREATE TABLE `AppNamespace` (
   `Id` int(10) unsigned NOT NULL AUTO_INCREMENT COMMENT '自增主键',
-  `Name` varchar(32) NOT NULL DEFAULT '' COMMENT 'namespace名字，注意，需要全局唯一',
+  `Name` varchar(500) NOT NULL DEFAULT '' COMMENT 'namespace名字，注意，需要全局唯一',
```
nobodyiam, line 64: Shall we also change the limitation in other places like [valdr.js](https://github.com/apolloconfig/apollo/blob/3440475ca0ad59fadc23f583c9fc36594dda1318/apollo-portal/src/main/resources/static/scripts/valdr.js#L66) so that users can create a namespace with length longer than 32?

nobodyiam (commented):
I think we also need to add the [upgrade scripts](https://github.com/apolloconfig/apollo/tree/master/scripts/sql/delta).

scripts/docker-quick-start/sql/apolloconfigdb.sql
klboke: I ignored that there is a layer of verification in the front end, and the input restrictions are considered to be redesigned so that the operation and maintenance personnel can adjust it. I'll close this pr first

PR #4452, opened 2022-07-08
feature:document support multiple version

### LINKED ISSUE #4327: document support multiple version
When user view the document https://www.apolloconfig.com/ , user may be only care about the document match their apollo's version.

For example, when user user's apollo version is 1.5.0, he/she may be confuse when see the feature above 1.5.0

**Describe the solution you'd like**
When someone use elasticsearch, its official site https://www.elastic.co/guide/en/elasticsearch/reference/7.17/index.html support choose the matched version of document.
image

Maybe we can add a button `version` on the top right of site?
image
and jump to that version's document which mark by the git tag.

**Additional context**
It's will improve the experience about the document, decrease the user's confuse.

falser101:
document support multiple version

Fixes #4327


falser101:
What kind of version style do you want？
@nobodyiam

nobodyiam:
Currently, [apolloconfig.com](apolloconfig.com) refers to the latest version of the [docs](https://github.com/apolloconfig/apollo/tree/master/docs) folder.
By supporting multiple versions of documentation, users could switch to different versions of the docs folder, e.g. when user chooses 1.9.2, then it should refer to the [1.9.2 tag](https://github.com/apolloconfig/apollo/tree/v1.9.2/docs).
I'm not sure how that could be achieved, but it should behave like this way.

falser101:
> 
Maybe we need to create version folder in docs and to put different versions of the files in the corresponding version folder.

falser101:
@Anilople

Anilople:
> > > 
> > 
> > 
> > Maybe we need to create version folder in docs and to put different versions of the files in the corresponding version folder. @nobodyiam
> 
> Currently, this seems the way to go. @Anilople do you have any suggestions?

That will consume large cost to maintain it.

Reference to different tag seems better.

Can we combine them together? i.e generate the version folder during CI rather than commit them to the git repo directly.

nobodyiam:
> generate the version folder during CI rather than commit them to the git repo directly.

It will be best to do so during CI as this is the most consistent way.
If we can't do so during CI, then we need to add a step to copy the docs to a version folder when releasing a new version.

falser101:
> It will be best to do so during CI as this is the most consistent way. If we can't do so during CI, then we need to add a step to copy the docs to a version folder when releasing a new version.

What should I do if we generate the version folder during CI? i don't kown how to do that

nobodyiam:
There is a proposal in docsify that apollo uses to render the sites: https://github.com/docsifyjs/docsify/issues/1289.
However, there doesn't seem to be any progress...

PR #4470, opened 2022-07-19
optimization apolloconfigdb-v190-v200-after.sql

klboke:

Create indexes to solve the problem of updating large tables，The effect is as follows
image


nobodyiam (approved):
LGTM

PR #4473, opened 2022-07-20
Optimize performance of '/apps/{appId}/envs/{env}/clusters/{clusterName}/namespaces' interface queries

klboke:

- As shown in the figure below, launching a 'findNamespaces' query will process each 'Namespace' in turn in the loop body, and when the 'Namespace ' is very large, e.g. up to 40, the total time taken is exaggerated to 7-8 seconds because of single-threaded serial execution.

-  Changing to multi-threaded processing of each 'Namespace' will have a very significant improvement on the 'findNamespaces' interface latency

> image
Timing diagram from @Anilople


apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/service/NamespaceService.java
```diff
@@ -165,20 +173,29 @@ public List<NamespaceBO> findNamespaceBOs(String appId, Env env, String clusterN
...
-        namespaceBOs.add(namespaceBO);
-      } catch (Exception e) {
-        LOGGER.error("parse namespace error. app id:{}, env:{}, clusterName:{}, namespace:{}",
-            appId, env, clusterName, namespace.getNamespaceName(), e);
-        throw e;
-      }
     }
-
+    try {
+      latch.await();
```
mghio, line 195: Is it better to use the method with timeout here? e.g. CountDownLatch.await(long timeout, TimeUnit unit)

klboke: I don't feel it's necessary

klboke: I mainly consider two points:

-  The remote interactions involved are all internal systems, and there is no unpredictable long blocking

- If you use  `CountDownLatch.await(long timeout, TimeUnit unit)` , Not very good to evaluate this timeout parameter, which will complicate the problem

mghio: Yes, the timeout is really not very good to set.

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/service/NamespaceService.java
```diff
@@ -165,20 +173,29 @@ public List<NamespaceBO> findNamespaceBOs(String appId, Env env, String clusterN
...
     for (NamespaceDTO namespace : namespaces) {
+      executorService.submit(() -> {
+        NamespaceBO namespaceBO;
+        try {
+          namespaceBO = transformNamespace2BO(env, namespace);
+          namespaceBOs.add(namespaceBO);
+        } catch (Exception e) {
+          LOGGER.error("parse namespace error. app id:{}, env:{}, clusterName:{}, namespace:{}",
+              appId, env, clusterName, namespace.getNamespaceName(), e);
+          throw e;
```
nobodyiam, line 187: I believe with the current change, the exception thrown here won't be aware in the calling thread.
We may use future.get to wait and retrieve the exception?

klboke: Indeed, I have modified and added test cases

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/service/NamespaceService.java
```diff
@@ -165,20 +173,32 @@ public List<NamespaceBO> findNamespaceBOs(String appId, Env env, String clusterN
...
-      }
+    }
+    try {
+      latch.await();
+    } catch (InterruptedException e) {
+      //ignore
     }
 
+    if(namespaceBOs.size() != namespaces.size()){
+       throw new RuntimeException("Parse namespace error");
```
Anilople, line 200: How about 

* encode size to error message
* find which namespace cannot get

for example

```suggestion
       throw new RuntimeException("Parse namespaces error, expect {} but get {}, cannot get those namespaces: {}");
```
This will give more information when trouble shoot.

klboke: This is a good idea

Anilople (dismissed):

apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/service/NamespaceService.java
```diff
@@ -165,20 +173,35 @@ public List<NamespaceBO> findNamespaceBOs(String appId, Env env, String clusterN
...
+    }
+    try {
+      latch.await();
+    } catch (InterruptedException e) {
+      //ignore
     }
 
+    if(namespaceBOs.size() != namespaces.size()){
+       throw new RuntimeException(String
+           .format("Parse namespaces error, expected: %s, but actual: %s, cannot get those namespaces: %s", namespaces.size(), namespaceBOs.size(), exceptionNamespaces));
```
nobodyiam, line 203: The original logic will throw the actual exception so that the user knows what's wrong. Is it better to preserve that logic? For example, we could use future.get to retrieve the exception occurred.

```java
    List<NamespaceBO> namespaceBOs = Collections.synchronizedList(new LinkedList<>());
    List<Future<?>> futures = new LinkedList<>();
    for (NamespaceDTO namespace : namespaces) {
      futures.add(executorService.submit(() -> {
        NamespaceBO namespaceBO;
        try {
          namespaceBO = transformNamespace2BO(env, namespace);
          namespaceBOs.add(namespaceBO);
        } catch (Exception e) {
          LOGGER.error("parse namespace error. app id:{}, env:{}, clusterName:{}, namespace:{}",
              appId, env, clusterName, namespace.getNamespaceName(), e);
          throw e;
        }
      }));
    }

    futures.forEach(future -> {
      try {
        future.get();
      } catch (ExecutionException | InterruptedException e) {
        Throwable exception = e instanceof ExecutionException ? e.getCause() : e;
        if (exception instanceof RuntimeException) {
          throw (RuntimeException) exception;
        } else {
          throw new RuntimeException(exception);
        }
      }
    });
```

klboke:
```java
LOGGER.error("parse namespace error. app id:{}, env:{}, clusterName:{}, namespace:{}",
              appId, env, clusterName, namespace.getNamespaceName(), e);
```
The original exception stack information will be recorded here

nobodyiam: I understand they are recorded in the server logs. What I propose is to throw the detailed exception message to the UI page so that the operating users could understand what's wrong, which is the original behavior.
The current error message `Parse namespaces error, expected: %s, but actual: %s, cannot get those namespaces: %s` could show the operating users some error details, but I have a concern that it might not be very clear.

klboke: I probably see what you mean, and the main points I took into account when making the changes here were.
- I am not very clear about the internal logic will report what possible exceptions, but when I simulated an exception, as a direct user it is difficult to do anything based on this exception, may be my simulation and the real difference
- the original will only encounter a Namespace problem to report an exception, now all the problems will be reported at once, will it be better?

nobodyiam: > I am not very clear about the internal logic will report what possible exceptions, but when I simulated an exception, as a direct user it is difficult to do anything based on this exception, may be my simulation and the real difference

The exceptions would be either timeout exceptions occurred in the portal side, or database access exceptions occurred in the admin service side. I suppose there is not much to do for a direct user, except that the user could report that to the admin users.
So my original idea was that we keep the logic the same as before - throw the actual exception and display it in the UI side, but I think the current logic is also ok.

> the original will only encounter a Namespace problem to report an exception, now all the problems will be reported at once, will it be better?

That I suppose is the same as the user can't really handle the errors, no matter it contains one error or multiple errors :-)

nobodyiam (approved):
LGTM

PR #4500, opened 2022-08-03
Sort Namespaces acquired by multiple threads

klboke:
Sort Namespaces acquired by multiple threads due to https://github.com/apolloconfig/apollo/pull/4473 causing the UI to display Namespaces out of order


apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/service/NamespaceService.java
```diff
@@ -202,7 +204,10 @@ public List<NamespaceBO> findNamespaceBOs(String appId, Env env, String clusterN
        throw new RuntimeException(String
            .format("Parse namespaces error, expected: %s, but actual: %s, cannot get those namespaces: %s", namespaces.size(), namespaceBOs.size(), exceptionNamespaces));
     }
-    return namespaceBOs;
+
+    return namespaceBOs.stream()
+        .sorted(Comparator.comparing(o -> o.getBaseInfo().getNamespaceName()))
+        .collect(Collectors.toList());
```
mghio, line 210: How about sorting by namespace id？because the query logic [here](https://github.com/apolloconfig/apollo/blob/master/apollo-biz/src/main/java/com/ctrip/framework/apollo/biz/repository/NamespaceRepository.java#L31) is to sort by namespace id. 
```suggestion
    return namespaceBOs.stream()
        .sorted(Comparator.comparing(o -> o.getBaseInfo().getId()))
        .collect(Collectors.toList());
```

nobodyiam: I think query by id is reasonable, which reflects the creation order.

klboke: If you use ID sorting, in different clusters, the order may be inconsistent, which is very unfriendly to users.

nobodyiam: Previously, the order of the namespaceBOs was the same as the namespaces returned from `namespaceAPI.findNamespaceByCluster(appId, env, clusterName)`. So I think it's better to keep the order whether we use parallel loading.

> If you use ID sorting, in different clusters, the order may be inconsistent, which is very unfriendly to users.

The logic of `namespaceAPI.findNamespaceByCluster(appId, env, clusterName)` is to sort the namespace by id asc, which reflects the creation order. So I think it should be natural for the users to understand.

klboke: for example:
image

- The above is the namespace order of the hk cluster. When a new sg cluster is created, the time to generate the namespace is uncontrollable, which leads to the inconsistent order of the namespaces of the hk and sg clusters. When there are many namespaces, the problem caused by the inconsistency of the order will be more prominent.

image

- However, on the UI interface, there is no place to display the creation time of the Namespace, so the user has no way of knowing why the Namespace is out of order.

- The above is the reason why I think it is better to use Namespace Name sorting to keep the order of each cluster consistent

nobodyiam: This is interesting. I didn't consider the new cluster scenario.
If we sort the namespaces by name, is it possible that the `application` namespace is not displayed at the top? I have a concern that this might make the users confused...
If we want to make the namespaces order consistent among clusters, how about we adjust the create cluster logic? e.g. create private namespaces based on the appnamespace id order

https://github.com/apolloconfig/apollo/blob/17f59619c026ba9525fccda341ab8605c291deaf/apollo-biz/src/main/java/com/ctrip/framework/apollo/biz/service/NamespaceService.java#L374-L390

klboke: - It seems to make a little more sense to maintain the previous sorting logic and keep the order of Namespace creation consistent across clusters.
- Before the main consideration has been how to solve the already existing disorder. Now it seems that direct ordering leads to more complicated problems, such as newly created Namespace may be ordered to an unknown place.

So, I agree that the ordering of ids should be maintained and has been changed, please review it again

nobodyiam (approved):
LGTM

PR #4509, opened 2022-08-11
1、fix transcation invalid

ksice:

Fix transaction invalid

sonar prompt bug is "assignNamespaceRoleToConsumer's" @Transactional requirement is incompatible with the one for this method.


ksice:

nobodyiam:
Thanks for the pull request. However, I'd like to learn more about what this change solves. Since there is no checked exception thrown for method `assignNamespaceRoleToConsumer`, I think adding `rollbackFor` or not is the same?

ksice:
> Thanks for the pull request. However, I'd like to learn more about what this change solves. Since there is no checked exception thrown for method `assignNamespaceRoleToConsumer`, I think adding `rollbackFor` or not is the same?

rollbackfor can not be added but the @transcation annotation should be added, otherwise the transaction of the "assignNamespace" method will be invalidated

nobodyiam:
Yes, but I think the `@Transactional` annotation is already added to `public List<ConsumerRole> assignNamespaceRoleToConsumer(String token, String appId, String namespaceName, String env)`?

ksice:
> Yes, but I think the `@Transactional` annotation is already added to `public List<ConsumerRole> assignNamespaceRoleToConsumer(String token, String appId, String namespaceName, String env)`?

yes,But what I mean is that this method assignNamespaceRoleToConsumer(String token, String appId, String namespaceName) needs to add @Transactional, @Transactional has several failure scenarios, one of which is
The @Transactional annotation does not take effect in similar method calls, so when assignNamespaceRoleToConsumer(String token, String appId, String namespaceName) is called, it will cause assignNamespaceRoleToConsumer(String token, String appId, String namespaceName, String env)
@Transactional fails, I find it not roll back when I locally called assignNamespaceRoleToConsumer(String token, String appId, String namespaceName) to throw an exception。
E.g
image

nobodyiam:
OK, then I think it's reasonable to add `@Transactional` to `assignNamespaceRoleToConsumer(String token, String appId, String namespaceName)`.

ksice:
> OK, then I think it's reasonable to add `@Transactional` to `assignNamespaceRoleToConsumer(String token, String appId, String namespaceName)`.

ok, I modified the commit

nobodyiam (approved):
LGTM

PR #4519, opened 2022-08-14
feature:check json value

### LINKED ISSUE #4142: 修改value时加上检测json是否正确的按钮
后台编辑value时，类似“ 检测隐藏字符”，假如录入的是JSON字符串，增加一个类似“检测JSON是否正确”的按钮，方便录入时主动检查JSON是否正确

image

AbnerHuang2:

feature:check json value

Fixes #4142 

## Brief changelog

to be more specific, I made a short video for it, It works like that.

https://user-images.githubusercontent.com/50099513/184520175-3b8f79d0-335a-44a5-9302-73f1be4d276c.mov


AbnerHuang2:

nobodyiam:
This feature looks great!
However, I'm wondering is it possible to display the json check button only for potential json values? As I could imagine there will be many check in the future, e.g. yaml check, XML check, etc. The edit modal will look very complicated if there are a lot of checks.

AbnerHuang2:
> This feature looks great! However, I'm wondering is it possible to display the json check button only for potential json values? As I could imagine there will be many check in the future, e.g. yaml check, XML check, etc. The edit modal will look very complicated if there are a lot of checks.

yeah, you are right, that will be terrible if there were so many check button.   but it is hard to show the check button if the value is not in JSON format, unless the value is in [] or {}, then I can show the check button. What do you think?

AbnerHuang2:
@nobodyiam  I have done this feature,  it works like that. 

https://user-images.githubusercontent.com/50099513/184789712-94f680e1-5bbc-4972-9d23-a8628ddca026.mov

apollo-portal/src/main/resources/static/views/component/item-modal.html
```diff
@@ -46,15 +46,22 @@ <h4 class="modal-title">
                     <label class="col-sm-2 control-label">{{'Component.ConfigItem.ItemValue' | translate }}</label>
                     <div class="col-sm-10" valdr-form-group>
                         <textarea id="valueEditor" name="value" class="form-control" rows="6" tabindex="2"
-                            ng-required="false" ng-model="item.value">
+                            ng-required="false" ng-model="item.value" ng-blur="check()">
```
nobodyiam, line 49: I think the experience of the `blur` event is not quite good, as it only displays the `check json` button after the focus leaves the textarea. However, that normally is when user clicks the submit button, so that the user would not even see the `check json` button.
Is it possible to display the `check json` button when the user is editing the value? e.g. to check whether the value is json text after user stops typing for several seconds?
Another potential area to improve is to display the `check json` button when the user just clicks the edit button but not editing anything in the textarea.

image

nobodyiam: BTW, would you please also update the [CHANGES.md](https://github.com/apolloconfig/apollo/blob/master/CHANGES.md)?

AbnerHuang2: > This feature looks great! However, I'm wondering is it possible to display the json check button only for potential json values? As I could imagine there will be many check in the future, e.g. yaml check, XML check, etc. The edit modal will look very complicated if there are a lot of checks.

from the begining, I thought you might want to show the check json button while the value is JSON , so I set the check rule is the value in {} or [] . but when user is typing , the value is not in {} or [] . because they type { or ] when they finish type. 
 if then , 
> Is it possible to display the check json button when the user is editing the value? e.g. to check whether the value is json text after user stops typing for several seconds?

the way you mentioned is the same as blur .
BTW,

> Another potential area to improve is to display the check json button when the user just clicks the edit button but not editing anything in the textarea.

as the rule (show check button when the value in {} or []) set before , the check button will not show in this scenario.

If you want show check button when user is typing , we will face the situation you metioned before , that is lots of check in defferent format.

so I am a little confused.  or I just replace the blur fuction  in other onchange function.

AbnerHuang2: > BTW, would you please also update the [CHANGES.md](https://github.com/apolloconfig/apollo/blob/master/CHANGES.md)?

I'll update the CHANGES.md later.

AbnerHuang2: > This feature looks great! However, I'm wondering is it possible to display the json check button only for potential json values? As I could imagine there will be many check in the future, e.g. yaml check, XML check, etc. The edit modal will look very complicated if there are a lot of checks.

from the begining, I thought you might want to show the check json button while the value is JSON , so I set the check rule is the value in {} or [] . but when user is typing , the value is not in {} or [] . because they type { or ] when they finish type.
if then ,

> Is it possible to display the check json button when the user is editing the value? e.g. to check whether the value is json text after user stops typing for several seconds?

the way you mentioned is the same as blur. and there is also comment to write after value. but in most time, after writing the value, user may click submit button directly, maybe onchange is better than blur. 
since the cost of check is tiny,  I'll replace blur with onchange . maybe it is better for UE(User Experience).

BTW,
> Another potential area to improve is to display the check json button when the user just clicks the edit button but not editing anything in the textarea.

as the rule (show check button when the value in {} or []) set before , the check button will not show in this scenario.

If you want show check button when user is typing , we will face the situation you metioned before , that is lots of check in defferent format. I think it may not an good idea.

nobodyiam: > since the cost of check is tiny, I'll replace blur with onchange

I'm not sure whether the onchange event could meet this requirement as it is triggered after the input loses focus?

> If you want show check button when user is typing , we will face the situation you metioned before , that is lots of check in defferent format. I think it may not an good idea.

I think the main issue that causes the problem is the item doesn't have format info. Maybe we could consider to add format info to the `item` entity, see https://github.com/apolloconfig/apollo/issues/1517#issuecomment-1119449593.

AbnerHuang2: > I think the main issue that causes the problem is the item doesn't have format info. Maybe we could consider to add format info to the item entity, see https://github.com/apolloconfig/apollo/issues/1517#issuecomment-1119449593.

I think it's a good idea. maybe we can use a selector to let user choose what the item value format exactly is, then we can show format hint. and in the way , we don't think onchange or blur anymore.

for example,
default display:
企业微信截图_18d53e90-3c01-45d7-af5f-5875ff494213

json format display:
企业微信截图_94a96cc3-96d0-4063-8a84-85da0309fc3b

AbnerHuang2: also, we can easy extend the selector to check other format,  and even custom check like git hook.

nobodyiam: In this case, I suppose we need to store the format in the `Item` table? Besides, we also need to take a look at the `Text` edit mode to see whether it is compatible or not.

AbnerHuang2: ok, maybe  it will be more complicated in this way. what's a touch issue! it has taken too much time in this issue.  maybe we should give the matter further thought.  in the long term,  if you want to support the item value format check more precise, store the format maybe is a good choice. but it should take more time to think about. 
at present, shall we just show check button in onchange event  to satisfy json format check?

AbnerHuang2: or  can I add your  wechat , if I have any idea, I can talk to you. and it may more effective for communicate.  if we reach a consensus, I can post and pr to deal with it. what do you think? if you don't wanna to be bothered, we can also talk in  the comment.

nobodyiam: Sounds good. You can find my email from my [profile page](https://github.com/nobodyiam) and tell me your contact info.

AbnerHuang2: well, nice. I'll send you later.  and I'll  review and test my code and commit tonight.

AbnerHuang2:
I have update CHANGES.md and add  json check when modal shown.  @nobodyiam

CHANGES.md
```diff
@@ -5,6 +5,7 @@ Release Notes.
 Apollo 2.1.0
 
 ------------------
+* [Add a potential json value check feature](https://github.com/apolloconfig/apollo/pull/4519)
```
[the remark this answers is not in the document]
AbnerHuang2: OK

nobodyiam (approved):
LGTM

PR #4571, opened 2022-09-15
Refactoring the message splicing of internal Exceptions

klboke:

Refactoring the message splicing of internal Exceptions

- As defined in AbstractApolloHttpException, the message splicing operation can be simplified when using

```java
public abstract class AbstractApolloHttpException extends RuntimeException{

  private static final long serialVersionUID = -1713129594004951820L;
  
  protected HttpStatus httpStatus;

  /**
   * When args not empty, use {@link com.google.common.base.Strings#lenientFormat(String, Object...)}
   * to replace %s in msgTpl with args to set the error message. Otherwise, use msgTpl
   * to set the error message. e.g.: 
   * <pre>{@code new NotFoundException("... %s ... %s ... %s", "str", 0, 0.1)}</pre>
   * If the number of '%s' in `msgTpl` does not match args length, the '%s' string will be printed.
   */
  public AbstractApolloHttpException(String msgTpl, Object... args){
    super(args == null || args.length == 0 ? msgTpl : Strings.lenientFormat(msgTpl, args));
  }
  
  public AbstractApolloHttpException(String msg, Exception e){
    super(msg,e);
  }

  protected void setHttpStatus(HttpStatus httpStatus){
    this.httpStatus = httpStatus;
  }

  public HttpStatus getHttpStatus() {
    return httpStatus;
  }
}

```


mghio (commented):
I think the format of `String.format` can be preserved, because it can warn when `%s` does not match the number of arguments. eg : `Too few arguments with format string (found: 0 , expected: 1)` .

klboke:
- @mghio when discussing this, does it mean choosing Strings.lenientFormat or String.format, whichever is more appropriate.

This pr (https://github.com/apolloconfig/apollo/pull/3999) has instructions

nobodyiam:
> I think the format of `String.format` can be preserved, because it can warn when `%s` does not match the number of arguments. eg : `Too few arguments with format string (found: 0 , expected: 1)` .

I think it's all right. As we could replace `Strings.lenientFormat` with `String.format` in `AbstractApolloHttpException` if we do want to warn the mismatched number of arguments.

nobodyiam (approved):
LGTM

mghio (approved):
LGTM

PR #4599, opened 2022-10-10
Fixes testUpdateBranchGrayRulesWithUpdateOnce

ZhewenFu:

Fix test `testUpdateBranchGrayRulesWithUpdateOnce`

Fixes the issue that causes flaky tests. The current test may lead to inconsistent ordering of stringified JSON objects and the test may fail sometimes. To fix the test and make it consistent, I select all the elements in the list and see whether the release history contains those elements. In this case, the ordering of the elements doesn't matter and won't cause failure.

## Brief changelog

Added containRule function.

### DIFF
--- a/apollo-biz/src/test/java/com/ctrip/framework/apollo/biz/service/NamespaceBranchServiceTest.java
+++ b/apollo-biz/src/test/java/com/ctrip/framework/apollo/biz/service/NamespaceBranchServiceTest.java
@@ -22,6 +22,9 @@ import com.ctrip.framework.apollo.biz.entity.Namespace;
 import com.ctrip.framework.apollo.biz.entity.ReleaseHistory;
 import com.ctrip.framework.apollo.common.constants.NamespaceBranchStatus;
 import com.ctrip.framework.apollo.common.constants.ReleaseOperation;
+import com.ctrip.framework.apollo.common.utils.GrayReleaseRuleItemTransformer;
+import com.ctrip.framework.apollo.common.dto.GrayReleaseRuleItemDTO;
+import java.util.Set;
 
 import org.junit.Assert;
 import org.junit.Test;
@@ -64,16 +67,17 @@ public class NamespaceBranchServiceTest extends AbstractIntegrationTest {
 
     namespaceBranchService.updateBranchGrayRules(testApp, testCluster, testNamespace, testBranchName, rule);
 
-    GrayReleaseRule activeRule = namespaceBranchService.findBranchGrayRules(testApp, testCluster, testNamespace,
-        testBranchName);
+    GrayReleaseRule
+        activeRule =
+        namespaceBranchService.findBranchGrayRules(testApp, testCluster, testNamespace, testBranchName);
 
     Assert.assertNotNull(activeRule);
     Assert.assertEquals(rule.getAppId(), activeRule.getAppId());
     Assert.assertEquals(rule.getRules(), activeRule.getRules());
     Assert.assertEquals(Long.valueOf(0), activeRule.getReleaseId());
 
-    Page<ReleaseHistory> releaseHistories = releaseHistoryService.findReleaseHistoriesByNamespace(testApp, testCluster,
-        testNamespace, pageable);
+    Page<ReleaseHistory> releaseHistories = releaseHistoryService.findReleaseHistoriesByNamespace
+        (testApp, testCluster, testNamespace, pageable);
 
     ReleaseHistory releaseHistory = releaseHistories.getContent().get(0);
 
@@ -81,15 +85,7 @@ public class NamespaceBranchServiceTest extends AbstractIntegrationTest {
     Assert.assertEquals(ReleaseOperation.APPLY_GRAY_RULES, releaseHistory.getOperation());
     Assert.assertEquals(0, releaseHistory.getReleaseId());
     Assert.assertEquals(0, releaseHistory.getPreviousReleaseId());
-    Assert.assertTrue(containRule(releaseHistory.getOperationContext(), rule.getRules()));
-  }
-
-  private boolean containRule(String context, String rule) {
-    String[] rules = rule.substring(2, rule.length() - 2).split(",");
-    for (int i = 0; i < rules.length; i++)
-      if (!context.contains(rules[i]))
-        return false;
-    return true;
+    Assert.assertTrue(containRule(releaseHistory.getOperationContext(),rule.getRules()));
   }
 
   @Test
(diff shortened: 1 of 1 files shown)


ZhewenFu:
Sorry I closed my previous PR by accident. The reason I didn't use [GrayReleaseRuleItemTransformer#batchTransformFromJSON](https://github.com/apolloconfig/apollo/blob/301d66dcc28b2d698eefed3ecc9aa3273afb9ddf/apollo-common/src/main/java/com/ctrip/framework/apollo/common/utils/GrayReleaseRuleItemTransformer.java#L35) is once we transform it into object, the object's `toString()` method still requires us to perform string manipulation.

zhewenfufufu:

zhewenfufufu:
recheck

apollo-biz/src/test/java/com/ctrip/framework/apollo/biz/service/NamespaceBranchServiceTest.java
```diff
@@ -64,25 +64,32 @@ public void testUpdateBranchGrayRulesWithUpdateOnce() {
 
     namespaceBranchService.updateBranchGrayRules(testApp, testCluster, testNamespace, testBranchName, rule);
 
-    GrayReleaseRule
-        activeRule =
-        namespaceBranchService.findBranchGrayRules(testApp, testCluster, testNamespace, testBranchName);
+    GrayReleaseRule activeRule = namespaceBranchService.findBranchGrayRules(testApp, testCluster, testNamespace,
```
nobodyiam, line 67: It's better not to involve unnecessary changes, as they would make it hard to distinguish what's actually changed.

apollo-biz/src/test/java/com/ctrip/framework/apollo/biz/service/NamespaceBranchServiceTest.java
```diff
@@ -64,25 +64,32 @@ public void testUpdateBranchGrayRulesWithUpdateOnce() {
...
+        testNamespace, pageable);
 
     ReleaseHistory releaseHistory = releaseHistories.getContent().get(0);
 
     Assert.assertEquals(1, releaseHistories.getTotalElements());
     Assert.assertEquals(ReleaseOperation.APPLY_GRAY_RULES, releaseHistory.getOperation());
     Assert.assertEquals(0, releaseHistory.getReleaseId());
     Assert.assertEquals(0, releaseHistory.getPreviousReleaseId());
-    Assert.assertTrue(releaseHistory.getOperationContext().contains(rule.getRules()));
+    Assert.assertTrue(containRule(releaseHistory.getOperationContext(), rule.getRules()));
```
nobodyiam, line 84: I understand the original `contains` check is not a good one, however, I would like to know the background behind this change. Is this test failed in some situations?
BTW, I would suggest parsing the string to object for comparison instead of simply comparing the string, as there is not much difference from the original check.

apollo-biz/src/test/java/com/ctrip/framework/apollo/biz/service/NamespaceBranchServiceTest.java
ZhewenFu: Thanks for pointing out! It was caused by autoformatting and now I have restored them to the original version.

apollo-biz/src/test/java/com/ctrip/framework/apollo/biz/service/NamespaceBranchServiceTest.java
ZhewenFu: > I understand the original `contains` check is not a good one, however, I would like to know the background behind this change. Is this test failed in some situations? BTW, I would suggest parsing the string to object for comparison instead of simply comparing the string, as there is not much difference from the original check.

* Is this test failed in some situations?

  Yes, the test failed in certain cases.  I detected that the test results are flaky using the NonDex tool for catching non-deterministic/flaky tests (https://github.com/TestingResearchIllinois/NonDex).

  You may reproduce the nondeterministic behavior using the following steps:
  Clone the repo and cd into it.
  `mvn install -am -pl RosettaCore -DskipTests`
  `mvn -pl RosettaCore test -Dtest=com.hubspot.rosetta.RosettaBinderTest#itHandlesAnnotatedOptionalGenericSetterSerialization - Confirm that the test passes as is.`
  `mvn -pl RosettaCore edu.illinois:nondex-maven-plugin:1.1.2:nondex -DnondexMode=ONE -Dtest=com.hubspot.rosetta.RosettaBinderTest#itBindsStoredAsJsonBeanCorrectly` - This will run the NonDex tool on the test. 

* Parsing the string to object for comparison instead of simply comparing the string

  Thanks for pointing this out! I have modified the code and currently, I am using [GrayReleaseRuleItemTransformer#batchTransformFromJSON](https://github.com/apolloconfig/apollo/blob/301d66dcc28b2d698eefed3ecc9aa3273afb9ddf/apollo-common/src/main/java/com/ctrip/framework/apollo/common/utils/GrayReleaseRuleItemTransformer.java#L35) to transfer the string to object and compare the element within the set. Let me know if you have any suggestions on this :)

nobodyiam:
The unit test failed, which seems related to the line of code

```java
Set<GrayReleaseRuleItemDTO> contextRules = GrayReleaseRuleItemTransformer.batchTransformFromJSON(context.substring(context.indexOf("["),context.lastIndexOf("]")+1));
```

ZhewenFu:
> The unit test failed, which seems related to the line of code
> 
> ```java
> Set<GrayReleaseRuleItemDTO> contextRules = GrayReleaseRuleItemTransformer.batchTransformFromJSON(context.substring(context.indexOf("["),context.lastIndexOf("]")+1));
> ```

Yeah that seems to be where the problem lies. I have changed it to transfer the context to a map struct to avoid hardcoding. Now it should work fine.

apollo-biz/src/test/java/com/ctrip/framework/apollo/biz/service/NamespaceBranchServiceTest.java
```diff
@@ -115,10 +122,24 @@ public void testUpdateBranchGrayRulesWithUpdateTwice() {
...
-    Assert.assertFalse(firstReleaseHistory.getOperationContext().contains(secondRule.getRules()));
-    Assert.assertTrue(secondReleaseHistory.getOperationContext().contains(firstRule.getRules()));
-    Assert.assertTrue(secondReleaseHistory.getOperationContext().contains(secondRule.getRules()));
+    Assert.assertTrue(containRule(firstReleaseHistory.getOperationContext(), firstRule.getRules()));
+    Assert.assertFalse(containRule(firstReleaseHistory.getOperationContext(), secondRule.getRules()));
+    Assert.assertTrue(containRule(secondReleaseHistory.getOperationContext(), firstRule.getRules()));
+    Assert.assertTrue(containRule(secondReleaseHistory.getOperationContext(), secondRule.getRules()));
+  }
+
+  private boolean containRule(String context, String rule) {
```
nobodyiam, line 131: How about we refactor this method a little bit to actually check whether the context contains some rule? It's not necessary for current test cases, but we could use that in the future.

```java
  private boolean containRules(String context, String rules) {
    Type grayReleaseRuleItemsType = new TypeToken<Map<String, Set<GrayReleaseRuleItemDTO>>>() {
    }.getType();
    Map<String, Set<GrayReleaseRuleItemDTO>> contextRulesMap = new Gson().fromJson(context, grayReleaseRuleItemsType);
    Set<GrayReleaseRuleItemDTO> ruleSet = GrayReleaseRuleItemTransformer.batchTransformFromJSON(rules);

    for (GrayReleaseRuleItemDTO rule : ruleSet) {
      boolean found = false;
      for (Set<GrayReleaseRuleItemDTO> contextRules : contextRulesMap.values()) {
        for (GrayReleaseRuleItemDTO contextRule : contextRules) {
          if (contextRule.toString().equals(rule.toString())) {
            found = true;
            break;
          }
        }
      }
      if (!found) {
        return false;
      }
    }

    return true;
  }
```

ZhewenFu: Yes, it looks excellent! This version does make more sense and makes the tests more consistent. Also thanks for fixing some style issues here :)

nobodyiam (approved):
LGTM

PR #4681, opened 2022-12-15
Unify the experience of using the portal UI

klboke:

Unify the experience of using the portal UI

The main changes are as follows：
- Adjust the page width of user management, system parameter management, import and export management to be consistent with other management pages
- Adjust the width of import and export management forms to be consistent with other forms


nobodyiam (approved):
LGTM

PR #4793, opened 2023-03-14
[codestyle] config idea default indent size to 2

Shoothzj:
## Brief changelog
Make the idea default indent size to 2, avoid config per developers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,6 @@ target
 .settings
 
 # Idea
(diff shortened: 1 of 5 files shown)


nobodyiam:
There is already a code style in https://github.com/apolloconfig/apollo/tree/master/apollo-buildtools/style. I suppose it's a common way not to commit the idea settings?

PR #4802, opened 2023-03-15
Fix the issue of the system permission management page retrieving non-existent users.

klboke:

This pr has two changes as follows.

- Retrieve user information from the user service API using the user ID obtained from the user permission table to avoid displaying non-existent user IDs on the page.
- Sort the retrieved user list to make the page display more user-friendly when there are a large number of users retrieved.

Trigger scenario description: We switched from the local user system to LDAP. When a user in LDAP is deleted, it is not deleted synchronously. It is expected that there should be no user ID that does not exist in the system. When I try to delete this user ID, the check for the existence of the deleted user is triggered, so this is very contradictory.


nobodyiam:
Please also fix the unit test.

nobodyiam (approved):
LGTM

PR #4803, opened 2023-03-15
Fix the problem that the deletion failure of the system rights management page does not prompt

klboke:
Fix the problem that the deletion failure of the system rights management page does not prompt

image


nobodyiam:
There is also a `toastr.warn` in `NamespaceRoleController.js`, does it needs to be fixed?

klboke:
> There is also a toastr.warn in NamespaceRoleController.js, does it needs to be fixed?

Yes, it has been fixed

nobodyiam (approved):
LGTM

PR #3542, opened 2021-02-06
Cache key normalize

Q-RK:
## What's the purpose of this PR

修复 config-service 开启缓存后对客户端 app.id 配置大小写敏感的问题

## Which issue(s) this PR fixes:
Fixes #3529

## Brief changelog

1. 在 ConfigServiceWithCache 中定义 LoadingCache 的装饰者类，负责处理key的大小写问题
2. ConfigServiceWithCache优先取大小写敏感的结果（兼容目前的逻辑）
3. 若区分大小的key获取的值为空，则key转为小写获取对应的值

Follow this checklist to help us incorporate your contribution quickly and easily:

- [X] Read the [Contributing Guide](https://github.com/ctripcorp/apollo/blob/master/CONTRIBUTING.md) before making this pull request.
- [X] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [X] Write necessary unit tests to verify the code.
- [X] Run `mvn clean test` to make sure this pull request doesn't break anything.

PR #4810, opened 2023-03-20
Improve the documentation for the config-service.cache.enabled configuration property.

klboke:
Improve the documentation for the config-service.cache.enabled configuration property.

### Motivation

- We encountered a situation in our production environment where, when caching was enabled, we were unable to retrieve the latest configuration due to a case mismatch between apollo.cluster keys.

### Possible solutions
If we want to enable in-memory caching and ensure that the loading behavior remains consistent with when it is not enabled, could we simply convert the cache key of Guava Cache to lowercase? Would this cause any additional problems? I'm not entirely sure. One thing to consider is that the code already handles namespaces in a similar way, so is this new issue consistent with that?


nobodyiam:
> If we want to enable in-memory caching and ensure that the loading behavior remains consistent with when it is not enabled, could we simply convert the cache key of Guava Cache to lowercase?

It's actually a known issue as is also recorded in the [apollo project](https://github.com/apolloconfig/apollo/projects/1). There was an attempt to fix this issue #3542 but not continued. I think we could still consider fixing it if there is an easier way.
image

klboke:
The worst part is that MySQL was not strictly set to match case sensitivity from the beginning, so now we have to do compatibility processing. I think we can merge this PR first, As for the compatibility processing issue, I will look into it later.

nobodyiam (approved):
LGTM

PR #4812, opened 2023-03-21
Refactor NotFoundException、BadRequestException

klboke:

- Give a unified description to exceptions of the same class


Shoothzj (dismissed):

nobodyiam (commented):
This is a good try to narrow down the messages. However, there seem 2 patterns to handle the exception:
1. BadRequestException.xxx
```java
throw BadRequestException.itemNotExists(itemId);
throw BadRequestException.namespaceNotExists();
```

2. new xxxNotFoundException
```java
throw new AppNotFountException(appId);
throw new ClusterNotFoundException(appId, clusterName);
```

Is there a special consideration for this? Or maybe we could unify them to one pattern?

klboke:
BadRequestException's HTTP response code is 400, while NotFoundException's HTTP response code is 404. This is their difference. When I was making modifications, I didn't pay attention to the specific logic of where each exception was thrown. Therefore, I kept the same semantics as before, mainly to make the exception information look uniform. Previously, various methods might have been used to concatenate exception information.

apollo-adminservice/src/main/java/com/ctrip/framework/apollo/adminservice/aop/NamespaceAcquireLockAspect.java
```diff
@@ -117,7 +117,7 @@ void acquireLock(long namespaceId, String currentUser) {
 
   private void acquireLock(Namespace namespace, String currentUser) {
     if (namespace == null) {
-      throw new BadRequestException("namespace not exist.");
+      throw BadRequestException.namespaceNotExists();
```
nobodyiam, line 120: Shall we also pass in the namespace info?

apollo-common/src/main/java/com/ctrip/framework/apollo/common/exception/BadRequestException.java
```diff
@@ -22,11 +22,97 @@
...
 
+  public static BadRequestException itemAlreadyExists(String itemKey) {
+    return new BadRequestException("item already exists for itemKey:%s", itemKey);
+  }
+
+  public static BadRequestException itemNotExists(long itemId) {
+    return new BadRequestException("item not exists for itemId:%s", itemId);
+  }
+
+  public static BadRequestException namespaceNotExists() {
```
nobodyiam, line 40: Maybe we could unify this method with `namespaceNotExists(String appId, String clusterName, String namespaceName)`?

apollo-adminservice/src/main/java/com/ctrip/framework/apollo/adminservice/aop/NamespaceAcquireLockAspect.java
klboke: Since we cannot obtain the namespace information from the method's definition, we cannot print it in the exception logs. This situation also occurs in other places.

apollo-common/src/main/java/com/ctrip/framework/apollo/common/exception/BadRequestException.java
klboke: When refactoring the log information, I preserved the previous code semantics because of these two exceptions with response codes of 400 and 404 respectively. So I did not merge them,

nobodyiam (approved):
LGTM

PR #4833, opened 2023-04-04
refactor(apollo logging): Simplify the default log path to /opt/logs

klboke:

refactor(apollo logging): Simplify the default log path to `/opt/logs`

### motivation

The default log path for all Apollo services is prefixed with 100003172, which seems to have no additional purpose but hinders quick locating of log files. It's time to remove it.


Shoothzj (dismissed):

nobodyiam:
There is an issue that the gc.log would conflict if they are running in one machine.
https://github.com/apolloconfig/apollo/blob/2800a00e928333d579fe1e78791c82c7cf352562/apollo-configservice/src/main/scripts/startup.sh#L95-L110

Shoothzj:
As I think, we can use a more reasonable name, like /opt/apollo/logs, not magic number

klboke:
@nobodyiam 
> There is an issue that the gc.log would conflict if they are running in one machine.

Adding prefix of ServiceName to gc.log path

apollo-adminservice/src/main/scripts/startup.sh
```diff
@@ -97,14 +97,14 @@ if [[ "$javaexe" ]]; then
     version=$(echo "$version" | awk -F. '{printf("%03d%03d",$1,$2);}')
     # now version is of format 009003 (9.3.x)
     if [ $version -ge 011000 ]; then
-        JAVA_OPTS="$JAVA_OPTS -Xlog:gc*:$LOG_DIR/gc.log:time,level,tags -Xlog:safepoint -Xlog:gc+heap=trace"
+        JAVA_OPTS="$JAVA_OPTS -Xlog:gc*:$LOG_DIR/$SERVICE_NAME-gc.log:time,level,tags -Xlog:safepoint -Xlog:gc+heap=trace"
```
nobodyiam, line 100: Is it better to have the name as `apollo-configservice.gc.log` versus `apollo-configservice-gc.log`?
```suggestion
        JAVA_OPTS="$JAVA_OPTS -Xlog:gc*:$LOG_DIR/$SERVICE_NAME.gc.log:time,level,tags -Xlog:safepoint -Xlog:gc+heap=trace"
```

klboke: `apollo-configservice.gc.log` looks better

nobodyiam (approved):
LGTM

PR #4889, opened 2023-05-27
Updated all dependencies to the latest versions

vishalsingh2972:


nobodyiam:
It's not recommended to explicitly set the version of a dependency. Typically, the version is defined within the [dependencyManagement](https://github.com/apolloconfig/apollo/blob/a1bf9ebaf4ae1880922f1019461816a0578f6b71/pom.xml#LL107C3-L107C23) section of the pom.xml file.

PR #5072, opened 2024-01-08
Add comment for clusters and UI display

BlackBear2003:

Add comment for clusters and UI display

Fixes #https://github.com/apolloconfig/apollo/issues/5067

## Brief changelog

it looks like: 

WX20240108-215908@2x
image

- [x] Read the [Contributing Guide](https://github.com/apolloconfig/apollo/blob/master/CONTRIBUTING.md) before making this pull request.


nobodyiam:
This looks great! Please find some comments below:
1. I think the comment is similar to the `remarks` of namespace, so it's better to use the same term, e.g. `Remarks` and `备注`
image
image

2. I think the comment should not be mandatory, as it's not a necessary information
3. When displaying the comment, it's not necessary to display the prefix `comment: `

BlackBear2003:
> I think the comment is similar to the remarks of namespace, so it's better to use the same term, e.g. Remarks and 备注

changed terms at front-end from 'comment' to 'remarks'.

I'm not sure should we change the sql and entity's field name? I see there is `comment` in AppNamespace, Commit and Item while there is not `remarks`. So I just use `remark` when UI display.

image
image

apollo-biz/src/main/java/com/ctrip/framework/apollo/biz/entity/Cluster.java
```diff
@@ -67,9 +70,17 @@ public void setParentClusterId(long parentClusterId) {
...
+  }
+
+  public void setComment(String comment) {
+    this.comment = comment;
+  }
+
   public String toString() {
     return toStringHelper().add("name", name).add("appId", appId)
-        .add("parentClusterId", parentClusterId).toString();
+        .add("parentClusterId", parentClusterId).add("comments", comment).toString();
```
nobodyiam, line 83:
```suggestion
        .add("parentClusterId", parentClusterId).add("comment", comment).toString();
```

nobodyiam (commented):
This looks fantastic! Below are a few comments for your consideration. Additionally, please update the CHANGES.md file to inform users about this new feature.

nobodyiam (approved):
LGTM

PR #5153, opened 2024-06-09
Fix npe of audit log get operator

BlackBear2003:

To fix existing NPE bug in audit logging

when there is no active span exists on request thread context, trying to get operator will cause NPE by continuously get().

- causing NPE codes
```java
      if (tracer != null) {
        return ((ApolloAuditTracer) tracer).scopeManager().activeSpan().operator();
      }
```

- prof by unit test

image

In some case this bug will appear: 

Set `audit.log.enabled=false` in Portal and `audit.log.enabled=true` in AdminService.

In this case, no activeSpan would be passed to AdminService, meanwhile AdminService would get operator through active span. And then cause NPE.

## Brief changelog

1. Fix this continuously get()
2. Reduce inelegant ways of exposing interfaces to the outside
3. Add unit test for ApolloAuditOperatorDefaultSupplier

- [ ] Read the [Contributing Guide](https://github.com/apolloconfig/apollo/blob/master/CONTRIBUTING.md) before making this pull request.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the retrieval of the active span's operator in audit logs, defaulting to "anonymous" if no active span exists.

- **Tests**
  - Added new test cases to ensure correct behavior when an active span is present or absent during audit operations.

### DIFF
--- a/apollo-audit/apollo-audit-impl/src/main/java/com/ctrip/framework/apollo/audit/context/ApolloAuditTracer.java
+++ b/apollo-audit/apollo-audit-impl/src/main/java/com/ctrip/framework/apollo/audit/context/ApolloAuditTracer.java
@@ -43,7 +43,7 @@ public class ApolloAuditTracer {
     this.operatorSupplier = operatorSupplier;
   }
 
-  public ApolloAuditScopeManager scopeManager() {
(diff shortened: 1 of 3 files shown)


nobodyiam (approved):
LGTM

PR #5325, opened 2025-01-28
feat: support incremental configuration synchronization client

jackie-coming:

https://github.com/apolloconfig/apollo/pull/5288
之前的讨论在这个pr，为什么会被关闭？ 我重新提交了pr，继续修复代码
Fixes #

## Brief changelog

XXXXX

- [ ] Read the [Contributing Guide](https://github.com/apolloconfig/apollo/blob/master/CONTRIBUTING.md) before making this pull request.

## Release Notes for Apollo Configuration Service v2.4.0

- **New Features**
  - Added support for incremental configuration synchronization
  - Clients can now receive partial configuration updates, reducing network overhead

- **Configuration Options**
  - New setting `config-service.incremental.change.enabled` introduced
  - Default value is `false`
  - When enabled, improves performance but may increase memory consumption

- **Performance**
  - Optimized configuration retrieval and synchronization process
  - Reduced network pressure on config service

- **Compatibility**
  - Supported on server versions 2.4.0+ and client versions 2.3.0+


jackie-coming:
@nobodyiam ：The logic for calculating configurationChanges should be placed in the IncrementalSyncConfigService. If the config cache is not enabled, data should be directly fetched from the database for comparison. If the config cache is enabled, data should be fetched from the cache for comparison, and the comparison results themselves can also be stored in the cache for reuse.


My question is, when incremental synchronization is enabled, is it better to use caching by default? Because not using caching will double the database I/O pressure compared to full synchronization.

nobodyiam:
@jackie-coming It seems you accidentally closed the previous pull request...

image

nobodyiam:
> @nobodyiam ：The logic for calculating configurationChanges should be placed in the IncrementalSyncConfigService. If the config cache is not enabled, data should be directly fetched from the database for comparison. If the config cache is enabled, data should be fetched from the cache for comparison, and the comparison results themselves can also be stored in the cache for reuse.
> 
> My question is, when incremental synchronization is enabled, is it better to use caching by default? Because not using caching will double the database I/O pressure compared to full synchronization.

In this context, I believe there are two variables: the size of the full configuration being pushed and the scale of the client base receiving the push.  
If the full configuration being pushed is small, then regardless of whether the client base is large or not, the benefits of incremental pushing are minimal.  
If the full configuration being pushed is large but the client base is small, then it might not be necessary to enable server-side caching; instead, directly fetching and calculating the incremental configuration from the database would suffice.  
If both the full configuration being pushed and the client base are large, then enabling both server-side caching and incremental pushing would yield significant benefits.

PR #5632, opened 2026-06-08
[codex] Add user access tokens

klboke:
## Summary
- Add portal-managed user access tokens for AI agents, stored as hashes with scoped operations, app/env/namespace constraints, expiry, rate limits, and audit metadata.
- Authenticate user tokens only for OpenAPI requests and route authorization through a dedicated user-token permission validator.
- Add user token whoami/current/capabilities endpoints with machine-readable OpenAPI action descriptions.
- Add user self-service token UI and super-admin token management UI for filtering, viewing, revoking, and deleting tokens.

## Validation
- mvn -pl apollo-portal -am -Dtest=UserTokenServiceTest,UserTokenControllerTest,UserTokenOpenApiControllerTest -Dsurefire.failIfNoSpecifiedTests=false test
- mvn -pl apollo-portal -am spotless:check
- git diff --check
- node --check apollo-portal/src/main/resources/static/scripts/controller/AdminUserTokenController.js
- node --check apollo-portal/src/main/resources/static/scripts/services/UserTokenService.js
- node JSON parse check for zh-CN/en i18n files
- mvn -pl apollo-assembly -am -DskipTests package
- Restarted local dev and verified portal=200 config=200 admin=200
- Runtime checked admin token API: tokenValue/tokenHash omitted, invalid status returns 400, unauthenticated requests redirect to /signin

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added user access tokens for OpenAPI automation and AI agents, supporting scope-based permissions (operations, apps, environments, namespaces).
  * New token management interface in Portal for creating, viewing, rotating, and revoking tokens.
  * Admin token management interface for user and token oversight.
  * Token capability discovery endpoints for understanding available operations and scopes.
  * Audit logging of token usage across OpenAPI requests.

* **Configuration**
  * Added configurable token expiration settings (default and maximum).
  * LDAP account status attribute mapping for disabled/locked user accounts.


nobodyiam:
Thanks for the update. I rechecked the latest head (`e715d352192b8313aadbcf13e2be1dbb2aacd4bc`) and the OpenAPI contract alignment now looks good: `UserTokenOpenApiController` implements the generated `UserTokenManagementApi` and returns generated `OpenUserToken*` models from the `apollo-openapi` v0.3.9 spec.

One small item still looks worth fixing before merge: `apollo-portal/src/main/resources/static/views/common/nav.html` still uses `href="{{ '/user-token.html' | prefixPath }}"` for the new user-token link. Please switch it to `ng-href="{{ '/user-token.html' | prefixPath }}"` so the browser never treats the raw Angular interpolation as a literal URL before compilation. The docs table comment appears fixed.

Note: this reply was generated and posted automatically by AI for initial triage; a maintainer will follow up if needed.

klboke:
Thanks for checking! Your comment was against `e715d352192b8313aadbcf13e2be1dbb2aacd4bc`. I pushed a follow-up commit afterwards (`cc2c672f5cff5f1cac49db0dbff12c1d940a8a68`) that switches the user-token nav link to `ng-href`.

Current code now has:

```html
<a ng-href="{{ '/user-token.html' | prefixPath }}" target="_blank" rel="noopener noreferrer">
```

I also checked the PR review threads and there are no unresolved threads left.

nobodyiam:
Thanks for the follow-up. I rechecked the latest head (`cc2c672f5cff5f1cac49db0dbff12c1d940a8a68`): the user-token nav link now uses `ng-href` and keeps `rel="noopener noreferrer"`, the `actions` docs table renders as a normal Markdown table, and the PR review threads are resolved.

I do not see a blocking issue from this pass. A maintainer can make the final approval/merge decision.

Note: this reply was generated and posted automatically by AI for initial triage; a maintainer will follow up if needed.

nobodyiam (changes requested):
Thanks for the update. I think this still needs changes before approval.

The main blocker is that user-token scope authorization does not appear to be applied consistently across the OpenAPI surface. A few examples:

1. InstanceController still only applies the config-hiding logic to UserInfoHolder.UserType.USER, while USER_TOKEN can reach instance responses that include configs without the same config-read/app/env/namespace scope checks. checkReleaseNamespaceReadAllowed also only handles CONSUMER explicitly.
2. NamespaceBranchController has similar gaps: branch reads can return NamespaceBO data, but the hide-config path only handles USER; branch mutation/operator helpers also do not handle USER_TOKEN consistently.
3. PermissionController exposes role/permission read APIs, but I do not see enforcement of the user-token action requirements advertised by the user-token capability catalog.

Please audit the OpenAPI controllers and make USER_TOKEN authorization explicit for every read/write surface, with tests covering denied scopes. Otherwise a scoped user token may be able to access data outside its intended app/env/namespace/action scope.

There is also a migration clarity issue in scripts/sql: the new delta files are under v240-v300. Since this change targets the 3.0.0 schema delta, the path should be v250-v300 unless there is a documented reason that v2.4.0 is the intended upgrade base. Keeping v240-v300 may mislead users about the 2.5.0 -> 3.0.0 migration path.

Please also avoid adding the new portal/controller/UserTokenController as a separate /user-tokens Web API surface. Apollo has already been moving both webapi and openapi under /openapi/v1, so these user-token management endpoints should follow the same contract-first path: update apollo-openapi, generate the API, and implement it in the existing portal management controller surface, preferably PortalManagementController. Otherwise this PR reintroduces a parallel controller/API style that the recent migration was trying to remove.

A few additional points to clarify or tighten:

- The ordering dependency between UserTokenAuthenticationFilter and ConsumerAuthenticationFilter should be made explicit. UserTokenAuthenticationFilter currently runs inside springSecurityFilterChain via addFilterBefore, while ConsumerAuthenticationFilter/PortalUserSessionFilter/UserTypeResolverFilter are ordered as servlet filters via FilterRegistrationBean. Since ConsumerAuthenticationFilter relies on the user-token request attribute being set first, please either keep the security-chain implementation and add a regression test/comment that guarantees springSecurityFilterChain runs before the OPEN_API_AUTH_ORDER filters, or move user-token authentication into the same ordered servlet-filter layer and remove the duplicate security-chain registration. Adding @Order alone would not make this dependency clear because the standalone registration is currently disabled.
- Please document what the apollo_pat_ prefix means and that it is the discriminator between user tokens and legacy consumer tokens.
- For denied read access with USER_TOKEN, returning empty lists/empty 200 responses may be confusing for agent/API clients. The existing hide-as-empty behavior makes sense for portal UI compatibility, but user-token callers would likely be better served by 403 so they can distinguish “no data” from “no permission”.
- The LDAP account-status mapping appears related because user-token authentication rejects disabled users, but it is a broader LDAP behavior change. Please call that relationship out in the PR description/tests, or consider splitting it if it is intended as a separate improvement.

klboke:
@nobodyiam Updated the PR based on the review feedback. Main changes in the latest pushed commit:

- Updated `apollo.openapi.spec.url` to `apollo-openapi` `v0.3.10` and kept the portal implementation aligned with the generated OpenAPI interfaces/models.
- Moved the user-token management implementation onto the generated `PortalManagementApi` path and removed the old handwritten portal controller path.
- Tightened `USER_TOKEN` permission checks across app, namespace, item, release, instance, branch, access-key, and permission endpoints so scoped tokens cannot read or operate on resources outside their scope.
- Added metadata-scope checks for `env.list` and `organization.list`; these now require a non-user-management operation scope and are reflected in the capability catalog.
- Made explicit-resource unauthorized reads return `403` for `USER_TOKEN`, while list-style endpoints still return the visible subset.
- Kept the SQL delta directory as `v250-v300` and fixed the delta headers to `from v2.5.0 to v3.0.0` so the path and header are self-consistent.
- Added/updated focused controller and permission tests for these cases.

Verified locally:

- `./mvnw -pl apollo-portal -am -Dtest=ItemControllerParamBindLowLevelTest,NamespaceControllerParamBindLowLevelTest,ClusterControllerParamBindLowLevelTest,AccessKeyControllerParamBindLowLevelTest,PermissionControllerTest,UserTokenPermissionValidatorTest,ReleaseBranchInstanceControllerTest,UserTokenOpenApiControllerTest,PortalUserTokenManagementControllerTest,AuthFilterConfigurationTest,EnvControllerTest,OrganizationControllerTest,AppControllerTest,AppControllerParamBindLowLevelTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvnw -pl apollo-portal -am spotless:check`
- `git diff --check`

Could you please take another look when you have time? Thanks.

klboke:
@nobodyiam 已根据最新 review 推了 0e46578b0，请再帮忙看一下。

本轮处理：
- `USER_TOKEN` 的 emergency publish 现在也会走 env 级别的 `isEmergencyPublishAllowed` 限制，避免绕过普通用户路径的限制。
- namespace list 接口恢复列表子集语义：当 user token 可见 namespace 全部被过滤时返回 `200 OK []`，不再返回 403。
- 修复 CI 里 `build (17)` 的兼容性失败：`EnvController` / `OrganizationController` 的 metadata scope 校验现在只对 `USER_TOKEN` 生效，避免影响 legacy consumer/client path。

验证：
- `./mvnw -pl apollo-portal -am -Dtest=ApolloOpenApiJavaClientCompatibilityTest,EnvControllerTest,OrganizationControllerTest,NamespaceControllerParamBindLowLevelTest,ReleaseBranchInstanceControllerTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvnw -pl apollo-portal -am spotless:check`
- `git diff --check`
- `./mvnw -B clean package jacoco:report -Dmaven.gitcommitid.skip=true`

当前 PR checks 已全部通过或 neutral。

nobodyiam (changes requested):
Thanks for the update. I rechecked the latest head (`0e46578b02e8b9ebd580741be10c5e4a8a091259`). Most of the previous review items look addressed: the SQL delta now uses `v250-v300`, the portal user-token endpoints were moved under `/openapi/v1` in `PortalManagementController`, the `apollo_pat_` prefix is documented, the filter ordering dependency is documented/tested, and the namespace-list subset behavior has been restored.

I still see two authorization blockers before approval:

1. `ReleaseController` still lets `USER_TOKEN` bypass the emergency-publish env gate. `createGrayDelRelease`, `createGrayRelease`, and `createRelease` all call `checkEmergencyPublishAllowedForUser`, but that helper still only applies `portalConfig.isEmergencyPublishAllowed(...)` when the auth type is `USER`. A scoped user token with release permission can therefore submit `isEmergencyPublish=true` even when the same portal user would be rejected. `NamespaceBranchController` was fixed to include `USER_TOKEN` in the same check; please apply the same fix in `ReleaseController` and add regression coverage for user-token normal/gray/gray-delete release requests.

2. App metadata read is not actually intersected with the owning user's current permissions. The docs say user-token permissions are evaluated from the intersection of the current user permissions and the token scope, but `UserTokenPermissionValidator#hasReadApplicationPermission` only checks token operation/app scope and does not consult the underlying user permission/role state. `AppController` then uses that result to filter `getAllApps()` for `/openapi/v1/apps`, `/apps/authorized`, and `/apps/by-self`; because an empty app scope means "all apps", a broad/default user token can list app metadata for apps the owning user no longer has access to. Please make app metadata read enforce the owning user's current app permissions as well, or reject unauthorized app scopes at token creation time, and add a test where the token scope allows an app but the owning user's current permission denies it.

nobodyiam (approved):
Thanks for the update. I rechecked the latest head (`c7f5c1e42e61a304c7620507b5e4018db4b6ce3c`).

The two remaining blockers from my previous review look addressed now:

- `ReleaseController` now applies the emergency-publish env gate to `USER_TOKEN` as well as `USER`, and the PR adds regression coverage for normal release, gray release, and gray-delete release requests.
- `UserTokenPermissionValidator#hasReadApplicationPermission` now intersects token operation/app scope with the owning user's current app roles/super-admin status, with tests covering both denied current-user permissions and unbounded token app scope.

The current checks are green/neutral and I do not see active unresolved review threads. I do not see remaining blockers from my side.

One non-blocking follow-up question: for the related CLI work in apolloconfig/apollo-cli#1, the current v0 CLI PR uses the existing consumer-token OpenAPI authorization model. Given this PR introduces user tokens for AI agents and automation, what do you think should be the preferred first-class auth mode for `apollo-cli`: keep consumer tokens for the initial v0 path for compatibility, or make user tokens the primary path once this lands? This would help align CLI onboarding, docs, and future auth UX.

klboke:
@nobodyiam Thanks for bringing this up. My preference is to make user tokens the preferred first-class auth mode for `apollo-cli` once a version containing this feature is available.

The main reason is delegation control. For AI agents and user-local automation, I think the safest default is that the human user explicitly decides what permission to delegate to the tool: which operations, apps, envs, clusters, and namespaces are allowed, with normal expiration, revocation, rotation, and audit behavior. With the latest changes in this PR, the effective authorization is also bounded by the current permissions of the owning user, so a user token should not become a way to retain access after the user loses the underlying app permission.

I still think consumer tokens should remain supported for compatibility and existing service-to-service integrations, especially for older Apollo deployments and initial CLI bootstrapping. But for CLI onboarding/docs after this lands, I would position user tokens as the recommended/default path, and consumer tokens as the legacy or compatibility path.

So for `apollo-cli`, I would suggest:

- support both auth modes where practical;
- document user token as the preferred mode for interactive users, AI agents, and personal automation;
- keep consumer token support for existing integrations and deployments that have not adopted user tokens yet.

That gives users an explicit, self-service way to decide whether and how much access they want to grant to their AI/CLI workflow, without forcing broader consumer-token credentials as the default.

PR #5634, opened 2026-06-24
Fix NPE in ConfigChangeContentBuilder.updateItem when the old value is null

vasiliy-mikhailov:
## Problem

`ConfigChangeContentBuilder.updateItem` compares the old and new item values with:

```java
if (!oldItem.getValue().equals(newItem.getValue())) {
```

When the old items value is `null` — e.g. a config item that previously had no value, or a null-valued item — this throws `NullPointerException`.

## Fix

Compare the values null-safely with `Objects.equals`.

## Test

Adds `testUpdateItemWithNullOldValue` and `testUpdateItemWithBothNullValues` to `ConfigChangeContentBuilderTest`. The null-old-value case throws `NullPointerException` before the change and passes after; the both-null case correctly records no change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved change detection to handle `null` values safely when comparing item updates, preventing unexpected errors and more accurately recording changes.

* **Tests**
  * Added coverage for update scenarios with `null` and non-`null` values.
  * Verified that no change is recorded when both values are `null`.

### DIFF
--- a/apollo-biz/src/main/java/com/ctrip/framework/apollo/biz/utils/ConfigChangeContentBuilder.java
+++ b/apollo-biz/src/main/java/com/ctrip/framework/apollo/biz/utils/ConfigChangeContentBuilder.java
@@ -23,6 +23,7 @@ import com.google.gson.GsonBuilder;
 import java.util.Date;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
 import org.springframework.beans.BeanUtils;
 
 public class ConfigChangeContentBuilder {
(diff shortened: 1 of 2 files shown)


vasiliy-mikhailov:

nobodyiam (approved):
Looks good to me. The update comparison is now null-safe with `Objects.equals`, and the added tests cover both the previous NPE case (`null` old value -> non-null value) and the both-null no-change case.

I also verified the focused `apollo-biz` test locally, and the current CI/checks plus review-thread state are clean.

PR #5635, opened 2026-06-24
Fix gray namespace value-length limit using the wrong namespace

vasiliy-mikhailov:
## Problem

`ItemService.getGrayNamespaceItemValueLengthLimit` resolves the parent namespace and then looks up its value-length limit — but passes the **gray** namespace instead of the parent:

```java
Namespace parentNamespace = namespaceService.findParentNamespace(grayNamespace);
if (parentNamespace != null) {
    int parentLimit = getItemValueLengthLimit(grayNamespace);   // should be parentNamespace
    if (parentLimit > grayNamespaceLimit) {
        return parentLimit;
    }
}
```

So a value-length override configured on the **parent** namespace is never applied to its gray namespaces — `getItemValueLengthLimit` just re-looks-up the gray namespaces own limit.

## Fix

Pass `parentNamespace` to `getItemValueLengthLimit` so the parent override is honored.

## Test

Adds `ItemServiceGrayNamespaceLimitTest`: with a parent namespace override of 50000 and a gray limit of 20000, `getGrayNamespaceItemValueLengthLimit` should return 50000. It fails before the change (returns the gray limit) and passes after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected gray-namespace item value length limit calculations to use the parent namespace’s limit when it is higher than the gray namespace’s limit.
  * Prevented valid item values from being incorrectly rejected due to an over-enforced gray-namespace limit.

* **Tests**
  * Added regression test coverage for gray-namespace limit handling, including both limit computation and end-to-end validation behavior.

### DIFF
--- a/apollo-biz/src/main/java/com/ctrip/framework/apollo/biz/service/ItemService.java
+++ b/apollo-biz/src/main/java/com/ctrip/framework/apollo/biz/service/ItemService.java
@@ -240,7 +240,7 @@ public class ItemService {
       int grayNamespaceLimit) {
     Namespace parentNamespace = namespaceService.findParentNamespace(grayNamespace);
     if (parentNamespace != null) {
-      int parentLimit = getItemValueLengthLimit(grayNamespace);
+      int parentLimit = getItemValueLengthLimit(parentNamespace);
       if (parentLimit > grayNamespaceLimit) {
         return parentLimit;
       }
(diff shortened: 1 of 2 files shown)


vasiliy-mikhailov:

nobodyiam (approved):
Thanks for the fix. The code now uses the resolved parent namespace when calculating the gray namespace value-length limit, so a parent namespace override can be honored as intended.

The added tests cover both the direct limit calculation and the `checkItemValueLength` validation path. The branch has been rebased onto the current master, CI is green, and I don't see any remaining code concerns.

PR #5336, opened 2025-02-11
Provide a new configfiles API to return the raw content of configuration files directly

Shawyeok:
## What's the purpose of this PR

In some application scenarios, if Apollo can provide an API to directly return the raw content of configuration files, users can reference a ConfigService URL directly without the need for an additional proxy URL. See details at https://github.com/apolloconfig/apollo/issues/5327

## Which issue(s) this PR fixes:
Fixes #5327

## Brief changelog

Provide a new configfiles API to return the raw content of configuration files directly

Follow this checklist to help us incorporate your contribution quickly and easily:

- [ ] Read the [Contributing Guide](https://github.com/apolloconfig/apollo/blob/master/CONTRIBUTING.md) before making this pull request.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Write necessary unit tests to verify the code.
- [ ] Run `mvn clean test` to make sure this pull request doesn't break anything.
- [ ] Update the [`CHANGES` log](https://github.com/apolloconfig/apollo/blob/master/CHANGES.md).

## Summary by CodeRabbit

- **New Features**
  - Introduced a "configfiles API" for retrieving raw configuration content directly.
  - Enhanced response capabilities for different configuration formats.

- **Documentation**
  - Updated user guides to clarify response formats for different namespace types and how to access raw configuration content.

- **Tests**
  - Added tests for the new raw configuration retrieval functionality and refined existing mock behaviors.

PR #5639, opened 2026-07-01
fix: enforce access key for raw /configfiles api

Shawyeok:
## Background
The raw `/configfiles` API was introduced in #5336. At that time, the access-key appId extraction logic was not updated for the new `/configfiles/raw/{appId}/...` route, so those requests could be checked against `raw` instead of the real application id.

## Summary
- extract the real appId for `/configfiles/raw/{appId}/...` access-key checks
- add regression coverage for raw config file URLs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app ID detection for an additional config file URL pattern, including “raw” config file requests.
  * Ensured requests under the new path format are parsed consistently with existing config endpoints.

* **Tests**
  * Added coverage for the new raw config file URL scenario to verify the extracted app ID is correct.

### DIFF
--- a/apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/util/AccessKeyUtil.java
+++ b/apollo-configservice/src/main/java/com/ctrip/framework/apollo/configservice/util/AccessKeyUtil.java
@@ -33,6 +33,7 @@ public class AccessKeyUtil {
   private static final String URL_SEPARATOR = "/";
   private static final String URL_CONFIGS_PREFIX = "/configs/";
   private static final String URL_CONFIGFILES_JSON_PREFIX = "/configfiles/json/";
+  private static final String URL_CONFIGFILES_RAW_PREFIX = "/configfiles/raw/";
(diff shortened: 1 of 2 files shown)


Shawyeok:
@nobodyiam @hezhangjian Please feel free to take a look.

nobodyiam:
Thanks for the fix. I checked the access-key extraction path and this looks aligned with the raw configfiles endpoint introduced in #5336: `/configfiles/raw/{appId}/...` is now handled before the broader `/configfiles/` branch, so the filter should look up the real appId instead of `raw`.

I also verified the added unit coverage with `AccessKeyUtilTest`; no blocking issue found from my pass.

Note: this reply was generated and posted automatically by AI for initial triage; a maintainer will follow up if needed.

nobodyiam (approved):
Looks good to me. I verified that `/configfiles/raw/{appId}/...` is handled before the broader `/configfiles/` branch, so access-key lookup now uses the real appId instead of `raw`.

I also checked the raw endpoint mapping and ran the focused `AccessKeyUtilTest`; no blocking issue found.

PR #5672, opened 2026-08-30
fix(portal): do not update unchanged items when revoking changes

### LINKED ISSUE #4659: revokeItem接口逻辑问题:删除item发布完成后，点击撤销修改，会导致多个配置被修改
- [x] 我已经检查过[discussions](https://github.com/ctripcorp/apollo/discussions)
- [x] 我已经搜索过[issues](https://github.com/ctripcorp/apollo/issues)
- [x] 我已经仔细检查过[FAQ](https://www.apolloconfig.com/#/zh/faq/common-issues-in-deployment-and-development-phase)

**描述bug**

简洁明了地描述一下bug

**复现**

通过如下步骤可以复现：

1. 删除namespace中间某一行
2. 点击“发布”，确定
3. 点击“撤销修改”，确定
4. 发现删除的某一行以下的所有行，配置的修改时间和修改人都被改成“当前时间”和“当前用户”。

**期望**

正常应该是无变化

**截图**

企业微信截图_8da001f6-7606-4e9d-b794-93834159150e
可以看到修改时间变成是一样的了

### 额外的细节和日志

- 版本：2.1.0
- 错误日志
- 配置：
- 平台和操作系统

henriquejsza:

Fix the `revokeItem` logic in `apollo-portal` so that revoking changes no longer produces spurious updates for items that were not actually modified.

**Root cause:** When revoking changes, `ItemService` compared each persisted item against the latest release and marked an item for update when either its value differed **or** its line number differed. Line numbers can shift whenever other items are added or removed (for example, deleting a middle row and publishing). As a result, every item below the shifted position was flagged for update even though its value was unchanged. On the server side this rewrote those items and reset their modification metadata (`DataChangeLastModifiedTime` / `DataChangeLastModifiedBy`) to the current time and current user, matching the behavior reported in the issue.

**Fix (minimal, value-only):** Restrict the update condition to value-only differences and ignore line-number shifts. Only items whose value actually changed are added to the update change set; pure line-number reordering no longer triggers an update, so unchanged items keep their original metadata.

```java
// before
} else if (!StringUtils.equals(oldItem.getValue(), value)
    || lineNum.get() != oldItem.getLineNum()) {
// after
} else if (!StringUtils.equals(oldItem.getValue(), value)) {
```

Fixes #4659

## Brief changelog

- `apollo-portal` `ItemService`: when revoking changes, only mark an item for update when its value differs from the released value; ignore line-number-only differences that occur when other items are added/removed. This prevents unchanged items from being rewritten and having their last-modified time/user reset.
- Added regression test `ConfigServiceTest#testRevokeItemShouldNotUpdateWhenOnlyLineNumbersDiffer`, which sets up persisted items sharing the same values as the latest release but with sparse/misaligned line numbers, invokes `revokeItem`, and asserts that the produced `ItemChangeSets` has 0 create, 0 update, and 0 delete items. This test fails (RED) against the previous condition and passes (GREEN) with the value-only fix.

- [x] Read the [Contributing Guide](https://github.com/apolloconfig/apollo/blob/master/CONTRIBUTING.md) before making this pull request.
- [x] Run `mvn spotless:apply` to format your code.
- [x] Update the [`CHANGES` log](https://github.com/apolloconfig/apollo/blob/master/CHANGES.md).

---

### Validation (JDK 17)

```bash
./mvnw -B spotless:apply
./mvnw -B clean package jacoco:report -Dmaven.gitcommitid.skip=true
```

Both complete with 0 failures / 0 errors. The `apollo-portal` `ConfigServiceTest` suite runs 9 tests, 0 failures, 0 errors, including the new regression test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Revoking configuration items no longer triggers unnecessary updates when only line numbers differ.
  * Preserves last-modified metadata for unchanged items.
  * Prevents unrelated item changes from being recorded when key/value content remains unchanged.

* **Tests**
  * Added coverage to verify that revocation produces no creates, updates, or deletes when only line positions differ.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,7 @@ Apollo 3.0.0
 * [Feature: support configurable OIDC username claim (`spring.security.oidc.user-id-claim-name`) for the Apollo user identity](https://github.com/apolloconfig/apollo/pull/5655)
 * [Fix: strict JSON/YAML well-formedness validation at the authoritative save path](https://github.com/apolloconfig/apollo/pull/5660)
 * [Change: upgrade Apollo server baseline to Spring Boot 4.1.1 and Spring Cloud 2025.1.3](https://github.com/apolloconfig/apollo/pull/5671)
+* [Fix: preserve last-modified metadata of unchanged items when revoking changes](https://github.com/apolloconfig/apollo/pull/5672)
 
 ------------------
 All issues and pull requests are [here](https://github.com/apolloconfig/apollo/milestone/18?closed=1)
--- a/apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/service/ItemService.java
+++ b/apollo-portal/src/main/java/com/ctrip/framework/apollo/portal/service/ItemService.java
@@ -241,8 +241,7 @@ public class ItemService {
             0 == deletedItemDto.getLineNum() ? lineNum.get() : deletedItemDto.getLineNum();
         changeSets.addCreateItem(buildNormalItem(0L, namespaceId, key, deletedItemDto.getType(),
             value, deletedItemDto.getComment(), newLineNum));
-      } else if (!StringUtils.equals(oldItem.getValue(), value)
-          || lineNum.get() != oldItem.getLineNum()) {
+      } else if (!StringUtils.equals(oldItem.getValue(), value)) {
         changeSets.addUpdateItem(buildNormalItem(oldItem.getId(), namespaceId, key,
             oldItem.getType(), value, oldItem.getComment(), oldItem.getLineNum()));
       }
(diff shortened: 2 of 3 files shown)


henriquejsza:

nobodyiam (approved):
The change matches the agreed scope for #4659. `revokeItem` now ignores synthetic line-number-only mismatches while retaining the existing value-change update path, and the regression test covers the sparse-line-number case.

I independently verified `ConfigServiceTest` on JDK 17: all 9 tests pass on `14d979a`, while restoring the previous condition makes the new regression test fail with 3 spurious updates. Required checks are green and there are no unresolved review threads. LGTM.

